PR #4, opened 2019-02-07
Support even more types

MattiasBuelens:
### Support type parameters in an extends clause
```typescript
interface Foo<T> {
  foo: T;
}
interface Bar {}
interface FooBar extends Foo<Bar> {} // <<<
```
### Support `this` type node
```typescript
interface Foo {
  foo: string;
}
declare class FooBuilder {
  foo(foo: string): this; // <<<
  build(): Foo;
}
```
**Drive-by fix:** the "implementation" for the class method with overrides had an incorrect type. This doesn't affect the generated type definitions (since the implementation's signature is always removed), but I thought it's better to give it a more "realistic" type. 😛

### DIFF
--- a/src/__tests__/testcases/generics/index.ts
+++ b/src/__tests__/testcases/generics/index.ts
@@ -12,10 +12,11 @@ interface L {}
 interface M {}
 interface N {}
 interface O {}
+interface P {}
 
 type Gen<T> = T;
 
-export interface I<T = A> {
+export interface I1<T = A> {
   a: T;
   b: Gen<B>;
 }
@@ -30,3 +31,5 @@ export class Cl<T = E> {
 export function fn<T = G>(g: T, h: Gen<H>) {}
 export type TyFn = <T = J>(j: T, k: Gen<K>) => L;
 export type TyCtor = new <T = M>(m: T, n: Gen<N>) => O;
+export interface I2 extends Gen<P> {
+}
--- a/src/__tests__/testcases/overrides/index.ts
+++ b/src/__tests__/testcases/overrides/index.ts
@@ -14,7 +14,7 @@ export default class Foo {
   method(c: C): D;
   method(e: E): F;
 
-  method(c: C | D): D | F {
-    throw c;
+  method(ce: C | E): D | F {
+    throw ce;
   }
 }
--- a/src/__tests__/testcases/shadowing/index.ts
+++ b/src/__tests__/testcases/shadowing/index.ts
@@ -1,1 +1,8 @@
-export { GenericInterface, GenericKlass, genericFunction, ConditionalInfer, Mapped } from "./exports";
+export {
+  GenericInterface,
+  GenericKlass,
+  genericFunction,
+  ConditionalInfer,
+  Mapped,
+  GenericType
+} from "./exports";
--- a/src/__tests__/testcases/shadowing/index.ts
+++ b/src/__tests__/testcases/shadowing/index.ts
@@ -4,5 +4,6 @@ export {
   genericFunction,
   ConditionalInfer,
   Mapped,
-  GenericType
+  GenericType,
+  GenericExtends
 } from "./exports";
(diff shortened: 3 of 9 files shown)


src/__tests__/testcases/overrides/index.ts
```diff
@@ -14,7 +14,7 @@ export default class Foo {
   method(c: C): D;
   method(e: E): F;
 
-  method(c: C | D): D | F {
-    throw c;
+  method(ce: C | E): D | F {
```
Swatinem: oh, interesting. Even though tests are excluded from the main tsconfig, this should have still given a local error, no?

src/__tests__/testcases/generics/index.ts
```diff
@@ -30,3 +31,5 @@ export class Cl<T = E> {
 export function fn<T = G>(g: T, h: Gen<H>) {}
 export type TyFn = <T = J>(j: T, k: Gen<K>) => L;
 export type TyCtor = new <T = M>(m: T, n: Gen<N>) => O;
+export interface I2 extends Gen<P> {
```
Swatinem: Nice!

I recently implemented type parameter shadowing.
Could you add this case to that test as well?

https://github.com/Swatinem/rollup-plugin-dts/blob/a831c2554330a4eb5846c0634a28f9bda2e58cbb/src/__tests__/testcases/shadowing/exports.ts#L19-L22

Swatinem:
And another question concerning `this`:
Does this also support the usage inside

```ts
function foo(this: SomeClass) {}
```

src/__tests__/testcases/overrides/index.ts
MattiasBuelens: `D` and `E` are empty interfaces so they are *structurally* identical, which is fine for the TypeScript compiler. 😄

src/__tests__/testcases/generics/index.ts
MattiasBuelens: I'm not 100% about how those shadowing tests work. Is this right?
```typescript
export interface GenericExtends<M = any, N = M> extends GenericInterface<M, N> {}
```

src/__tests__/testcases/shadowing/index.ts
```diff
@@ -1 +1,8 @@
-export { GenericInterface, GenericKlass, genericFunction, ConditionalInfer, Mapped } from "./exports";
+export {
+  GenericInterface,
+  GenericKlass,
+  genericFunction,
+  ConditionalInfer,
+  Mapped,
+  GenericType
```
MattiasBuelens: This one was missing, so it wasn't actually being tested. Luckily, it works fine. 😅

src/__tests__/testcases/type-this/expected.d.ts
```diff
@@ -0,0 +1,5 @@
+declare class Foo {
+    a: this;
+}
+declare function thisType(this: Foo): void;
```
MattiasBuelens: > 
> 
> And another question concerning `this`:
> Does this also support the usage inside
> 
> ```ts
> function foo(this: SomeClass) {}
> ```

As you can see here: that works indeed. 😄

src/__tests__/testcases/generics/index.ts
MattiasBuelens: I've added (what I think is) a test for this in bed58e741f4413ea08ce6732c5e20ff2dac42228. If that's not correct, let me know. 🙂

src/__tests__/testcases/overrides/index.ts
Swatinem: Haha, you are absolutely right. Reminds me how I miss nominal typing :-D

Swatinem:
perfect, thanks :-)
I will publish tomorrow as soon as I am on my laptop.

Swatinem:
published `v0.10.0` with your changes

MattiasBuelens:
@Swatinem It looks like you forgot to push the `0.10.0` tag [to GitHub](https://github.com/Swatinem/rollup-plugin-dts/tags). Nothing critical, just thought I'd let you know. 😉

Swatinem:
meh forgot the `--follow-tags` :-D

Issue #24, opened 2019-05-30
"const" declarations from .d.ts modules should be marked with "declare"


aleclarson:
Repro: https://github.com/aleclarson/repro/tree/rollup-dts-1

I also tested with `.ts` modules, and that works fine. The problem is only with `.d.ts` modules.

cspotcode:
@Swatinem shouldn't the plugin add the `declare` modifier?  Is that too messy to implement?  As far as I can tell, it is implied by `export`, so whenever the input has `export`, the output should have `declare`.

As I understand it, rollup-plugin-dts is moving the `export` modifier to a different statement, which means it is changing and breaking the input.

Swatinem:
I think by now I have code that properly adds the missing attributes. Do you have a specific snippet that is failing?

cspotcode:
@Swatinem sure, here's a minimal reproduction: https://github.com/cspotcode/repros/tree/Swatinem/rollup-plugin-dts/24

Thanks for taking a look.

Swatinem:
Ah thanks for the repro case, this was indeed straight forward to fix ;-)

Issue #15, opened 2019-04-15
Plugin interferes with Rollup watch


robhicks:
I tried the plugin over the weekend. It works well except I cannot get rollup watch to work with it.  

Please let me know what I'm doing wrong: 
https://bitbucket.org/robhicks55/rollup-plugin-dts-watch/src/master/

Swatinem:
I don’t think you are doing anything wrong, rather the plugin is fundamentally incompatible with `rollup-watch`.
I basically use the TS compiler internally, which itself compiles/typechecks the whole project at once. Making it work with a watcher would probably involve some work inside the plugin.

Swatinem:
`1.0.0-alpha.0` should work. at least it does not do any internal caching itself.

PR #25, opened 2019-05-30
test: ambient const/function

aleclarson:
Related: #24

This does *not* include a fix. 😢

### DIFF
--- a/tests/testcases.test.ts
+++ b/tests/testcases.test.ts
@@ -101,7 +101,7 @@ describe("rollup-plugin-dts", () => {
         tsconfig: path.join(TESTCASES, "tsconfig.json"),
       };
       const rollupOptions: InputOptions = {
-        input: "index.ts",
+        input: name.startsWith("ambient-") ? "index.d.ts" : "index.ts",
       };
       const meta: Meta = {
         skip: false,
--- a/tests/testcases/ambient-const/expected.d.ts
+++ b/tests/testcases/ambient-const/expected.d.ts
@@ -1,0 +1,3 @@
+declare const tuple: [number, string];
+declare const unique: unique symbol;
+export { tuple, unique };
--- a/tests/testcases/ambient-const/index.d.ts
+++ b/tests/testcases/ambient-const/index.d.ts
@@ -1,0 +1,2 @@
+export const tuple: [number, string];
+export const unique: unique symbol;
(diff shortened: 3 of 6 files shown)


aleclarson:
I've also made it so the `intl-codegen` test is skipped by default, since it requires a symlink to pass.

Swatinem:
@aleclarson thanks for the report and the testcases. :-)

So far I haven’t very much tested consuming *handwritten* `.d.ts` files.
The `.d.ts` files generated by `tsc` should all have `declare` modifiers on them, except for `export default`, which I work around in a quite ugly way

https://github.com/Swatinem/rollup-plugin-dts/blob/191e42805662e66be867968285d21d8ead95d017/src/compiler.ts#L178-L182

So my question is, are you sure that the `.d.ts` files are actually valid? Running `tsc` on them with `skipLibCheck: false` yields the following error at least:

```
error TS1183: An implementation cannot be declared in ambient contexts.

6 export function foo() {}
                        ~
```

aleclarson:
Fixed the `ambient-function` test case!



> So far I haven’t very much tested consuming _handwritten_ `.d.ts` files.

Ah that's too bad. The `.d.ts` modules I have are definitely correct. Looks like we need to replace the `export` with a `declare` instead of merely stripping the `export` keyword.

Swatinem:
I think this is a bit related to the way I use the typescript compiler.

I basically do an ahead-of-time compile, because it is the easiest to implement. However it causes a few problems, one of them is performance, and also #15.

I think I can solve those, and also this problem when I switch over to using a different compiler mode of typescript, which hopefully lets me do all of this stuff…

However its quite complex to implement -_- as all these APIs are not quite that well documented.

How urgent is this for you? And are these files indeed *hardwritten*? In which case you can either just add the `declare`, or if its an external `@types`, just add it to `external` to not process it at all.

aleclarson:
No hurry. I can just skip bundling altogether for now. Thanks!

aleclarson:
I decided to stop using `.d.ts` modules in most of `react-spring` and then add an explicit `declare` in the few remaining `.d.ts` modules, and this has worked nicely.

Another thing to note: This package seems to prepend the `declare` keyword to every `type` declaration, and that's not necessary (though, it doesn't hurt either).

Swatinem:
> This package seems to prepend the `declare` keyword to every `type` declaration

Well thats typescripts own doing :-)

Swatinem:
going to close this as per https://github.com/Swatinem/rollup-plugin-dts/issues/24#issuecomment-498238530

PR #32, opened 2019-06-06
feat: support ".ts" modules

### LINKED ISSUE #31: Support non-ambient ".ts" modules
I'm curious why support for non-ambient (eg: `.ts` not `.d.ts`) definitions was removed.

I don't want this plugin to emit anything other than the `.d.ts` bundle, but it would be nice to avoid generating intermediate `.d.ts` modules. <s>Also, with `ts-loader` and `awesome-typescript-loader`, it's not possible to generate `.d.ts` modules if you want to use a forked process for type checking (eg: with [`fork-ts-checker-webpack-plugin`](https://www.npmjs.com/package/fork-ts-checker-webpack-plugin)).</s>

To be clear, I don't want this plugin to perform type checking, either.

aleclarson:
This compiles ".ts" modules into ".d.ts" modules in-memory and forwards them to Rollup.

This removes the need for intermediate files.

Fixes #31

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -1,6 +1,6 @@
 import * as ts from "typescript";
 import * as path from "path";
-import { PluginImpl, SourceDescription, InputOptions } from "rollup";
+import { PluginImpl, SourceDescription } from "rollup";
 import { Transformer } from "./Transformer";
 import { NamespaceFixer } from "./NamespaceFixer";
 
@@ -29,15 +29,12 @@ function getCompilerOptions(input: string): ts.CompilerOptions {
     return {};
   }
   return options;
-};
+}
 
-const createProgram = ({ input }: InputOptions) => {
-  if (typeof input !== "string") {
-    throw new TypeError('"input" option must be a string');
-  }
-  input = path.resolve(input);
+const createProgram = (main: string) => {
+  main = path.resolve(main);
   const compilerOptions: ts.CompilerOptions = {
-    ...getCompilerOptions(input),
+    ...getCompilerOptions(main),
     // Ensure ".d.ts" modules are generated
     declaration: true,
     // Skip ".js" generation
@@ -52,7 +49,7 @@ const createProgram = ({ input }: InputOptions) => {
     preserveSymlinks: true,
   };
   const host = ts.createCompilerHost(compilerOptions, true);
-  return ts.createProgram([input], compilerOptions, host);
+  return ts.createProgram([main], compilerOptions, host);
 };
 
 // Parse a TypeScript module into an ESTree program.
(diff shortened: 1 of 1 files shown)


src/index.ts
```diff
@@ -1,13 +1,86 @@
...
+const dts = ".d.ts";
+const tsx = /\.tsx?$/;
+
+const formatHost: ts.FormatDiagnosticsHost = {
+  getCurrentDirectory: () => ts.sys.getCurrentDirectory(),
+  getNewLine: () => ts.sys.newLine,
+  getCanonicalFileName: ts.sys.useCaseSensitiveFileNames ? f => f : f => f.toLowerCase(),
+};
+
+const getCompilerOptions = (input: string): ts.CompilerOptions => {
```
```suggestion
function getCompilerOptions(input: string): ts.CompilerOptions {
```
my personal preference is to have function declarations on the toplevel

src/index.ts
```diff
@@ -1,13 +1,86 @@
...
+  if (errors.length) {
+    console.error(ts.formatDiagnostics(errors, formatHost));
+    return {};
+  }
+  return options;
+};
+
+const createProgram = ({ input }: InputOptions) => {
+  if (typeof input !== "string") {
+    throw new TypeError('"input" option must be a string');
```
Swatinem: well this fails the multiple entrypoints test

src/index.ts
```diff
@@ -32,6 +105,43 @@ const plugin: PluginImpl<{}> = () => {
...
+        return null;
+      }
+      const source = program.getSourceFile(id);
+      if (!source) {
+        return null;
+      }
+      if (id.endsWith(dts)) {
+        return transformFile(source);
+      }
+      const ambientId = id.replace(tsx, dts);
```
Swatinem: The `ambient` naming confuses me a bit, rather name it `definitionId` / `definitionSource`.

Swatinem: or rather `declarationX`, to better reflect how its named in TS

aleclarson: 👍

src/index.ts
aleclarson: I will fix by calling `ts.createProgram` once per entry point. Sound good?

aleclarson:
> lazily call `createProgram` inside of `load` when you hit a `.tsx?` file

Instead, I'm going to only use `createProgram` when the entry point is a `.tsx?` file :)

src/index.ts
aleclarson: We could add a way to configure the plugin to use a single `ts.Program` for all the entry points, as long as they're all `.ts` modules (and not `.d.ts` modules). It just needs to know which `tsconfig.json` to use.

aleclarson:
LGTM 👍

Swatinem:
> The failing tests are now unrelated to this PR.

They are related, since you now return `.ts` instead of `.d.ts` from the `resolveId` hook.

I can take care of this tomorrow when I’m at my work machine.

PR #34, opened 2019-06-07
test: export all

aleclarson:
This test fails 😢


Swatinem:
I think this is an upstream issue, I filed it here: https://github.com/rollup/rollup/issues/2904

luckily @lukastaegert is super responsive with all rollup issues :-) :heart:

Issue #35, opened 2019-06-11
Make `.ts` mode compatible with `rollup-watch`


Swatinem:
CC @aleclarson 

Hm, I completely forgot about #15 when I accepted #32.

Now (again), it uses `createProgram` ahead of time and queries that `program` instance inside the `load` hook. But the `program` instance is never invalidated based on *watch* feedback.

Thats also what I always meant when I said that using the typescript api for anything more than very trivial usecases is quite hard…

aleclarson:
Yikes, I can't find any docs on integrating with TypeScript's watch-related API.

After some digging, I think [`ts.createWatchProgram`](https://github.com/microsoft/TypeScript/blob/d023427226ca3c7e900ddc65606a56e9f11802eb/src/compiler/watch.ts#L638-L646) is the right place to start.

weeebdev:
Any updates on that?

Swatinem:
I very much discourage using `.ts` mode. And I won’t be pouring in an unreasonable amount of effort into this.

Swatinem:
I believe this was just fixed in #249

PR #36, opened 2019-06-15
test: "export const" in .ts module

aleclarson:
This test fails with the following output:
```diff
- /// <reference types="react" />
- declare type JSXElements = keyof JSX.IntrinsicElements;
- declare const elements: JSXElements[];
+ export declare type JSXElements  JSX.IntrinsicElements;
+ export declare const elements: Jnts[];
  export { JSXElements, elements };
```


aleclarson:
Looks like TypeScript is giving us an incorrect AST. 🤔

aleclarson:
Looks like two bugs with one stone!

- Use `input.getFullText` instead of `input.getText` (which omits "comment pragmas")

- Comment pragmas don't get AST nodes (solution needed)

aleclarson:
Looks like two bugs with one stone!

- Use `input.getFullText` instead of `input.getText` (which omits "comment pragmas")
  (fixed in https://github.com/Swatinem/rollup-plugin-dts/pull/36/commits/936446aa5fc41b5be2e30c37a0fe33b18f195b97)

- Rollup omits comment pragmas, because they use `//` syntax
  (not fixed yet, but I opened this: https://github.com/rollup/rollup/issues/2943)

aleclarson:
Hmm, looks like we'll need to prepend the removed comment pragmas post hoc.

> Preserving certain comments would be a non-trivial change, especially since by default, acorn does not help us at all in associating comments with Nodes so everything needs to be done by hand.

aleclarson:
Alright, the test is passing now! 🎉 

With https://github.com/Swatinem/rollup-plugin-dts/pull/36/commits/ca5ba31ed34795ae1578244669890b5bfade3451, any `reference` directives are collected in the `transformFile` function, and then they're re-added during the `renderChunk` step.

tests/testcases/export-ts/index.ts
```diff
@@ -0,0 +1,137 @@
+export type JSXElements = keyof JSX.IntrinsicElements;
```
Swatinem: please give the testcase a more meaningful name, such as `preserve-tripleslash-references`

src/index.ts
```diff
@@ -56,6 +46,35 @@ const plugin: PluginImpl<{}> = () => {
     return { source, program };
   }
 
+  // Parse a TypeScript module into an ESTree program.
+  const references = new Set<string>();
```
Swatinem: I don’t really have a better idea right now, just keep in mind that introducing more *global state* makes it harder to fix watch mode (#35)

src/index.ts
```diff
@@ -6,24 +6,14 @@ import { Transformer } from "./Transformer";
...
-  // hm, typescript generates `export default` without a declare,
-  // but rollup moves the `export default` to a different place, which leaves
-  // the function declaration without a `declare`.
-  // Well luckily both words have the same length, haha :-D
-  let code = input.getText();
-  code = code.replace(/(export\s+)default(\s+(function|class))/m, "$1declare$2");
-  for (const fixup of fixups) {
-    code = code.slice(0, fixup.range.start) + fixup.identifier + code.slice(fixup.range.end);
-  }
+const toArray = <T>(value: T): T extends ReadonlyArray<any> ? T : [T] =>
```
```suggestion
function toArray<T>(value: T | ReadonlyArray<T>): Array<T> {
  return Array.isArray(value) ? value : [value];
}
```

Well this is just personal preference, but since I now have external contributors 🎉 , it might be a good idea to set up some linting:

* prefer function declarations over arrows on the toplevel
* avoid `any` as much as possible (you cast things a couple of times)
* prefer generics notation over `T[]` (you actually made this a one-element tuple, not sure if that was intentional?)

src/index.ts
```diff
@@ -6,24 +6,14 @@ import { Transformer } from "./Transformer";
...
+  Array.isArray(value) ? (value as any) : [value];
 
-  return { code, ast };
-}
+type PragmaMap = Map<"reference", Pragma | Pragma[]>;
+type Pragma = {
```
Swatinem: also related to linting: My personal preference is to use `interface` declarations

src/index.ts
```diff
@@ -56,6 +46,35 @@ const plugin: PluginImpl<{}> = () => {
...
   }
 
+  // Parse a TypeScript module into an ESTree program.
+  const references = new Set<string>();
+  function transformFile(input: ts.SourceFile): SourceDescription {
+    let code = input.getFullText();
+
+    // Collect any reference directives
+    const pragmas: PragmaMap = (input as any).pragmas;
+    if (pragmas && pragmas.has("reference")) {
```
Swatinem: since you check this here, is your type declaration really correct? (nullability?)
Also, your manual type declaration plus the `as any` cast suggests that this is not officially documented TS API?

A quick check with https://astexplorer.net/#/gist/577600e28bd89938dc16f5bf445980d4/0213bb7d7515552ab575dabaec61561145080b5d showed me that the sourceFile has a `typeReferenceDirectives` array. Maybe using that would be a better idea?
(these is also `libReferenceDirectives`)

That way, you can really re-create the tripleslash references manually.
Regarding the issue you opened for `rollup`, I think it is *guaranteed* to remove those comments, right? So we will not end up with a case where we end up with duplicated comments, right?

src/index.ts
aleclarson: > (you actually made this a one-element tuple, not sure if that was intentional?)

It was intentional, since we know the array only has one value when a non-array is passed, but it's largely inconsequential. :)

src/index.ts
aleclarson: Nice, I didn't see the `typeReferenceDirectives` property before.



> I think it is _guaranteed_ to remove those comments, right? So we will not end up with a case where we end up with duplicated comments, right?

I don't think so. I see a preserved directive in the `.d.ts` bundle of `@react-spring/web`. 😕 
Looks like we'll need a test case for that.

aleclarson: Though, that might be because I'm using the `renderChunk` hook to prepend the comment pragmas. There might be a better place to do that?

Swatinem: > Looks like we'll need a test case for that.

yes, that was also my thought to test that comments that are attached to items are correctly preserved / removed. that pragma is kind of weird, according to astexplorer its attached to the export statement, but that does not really make sense IMO.

src/index.ts
```diff
@@ -49,17 +38,11 @@ const plugin: PluginImpl<{}> = () => {
...
-    let code = input.getFullText();
-
-    // Collect any reference directives
-    const pragmas: PragmaMap = (input as any).pragmas;
-    if (pragmas && pragmas.has("reference")) {
-      toArray(pragmas.get("reference")!).forEach(({ range }) => {
-        references.add(code.slice(range.pos, range.end));
-      });
-    }
+    input.typeReferenceDirectives.forEach(ref => {
```
Swatinem: I prefer `for … of` over `forEach` :-)

Swatinem:
Thank you very much :-)
I will publish this tomorrow, and possibly play a bit more with comment attachment when I find some time.

PR #47, opened 2019-07-19
Failing test case for inspect.custom symbol

andykais:
I have a library where this is occurring. This is the minimal reproduction.

The `import { inspect } 'util';` is missing from the dts rollup


Swatinem:
Thanks for the testcase :-)

I have fixed this in `master`, but I’m unable to publish right now, as I need a new 2FA device. So please wait a few more days :-)

andykais:
awesome, no rush on the fix. Thank you very much!

andykais:
Thank you!

PR #48, opened 2019-07-22
Failing test case for exporting multiple variables

andykais:
There may be more than one thing going wrong in this test case. At the very least, some of the variables are not being exported.

I was experiencing this in one of my repos and this is my minimal repro.

### DIFF
--- a/tests/testcases/export-multiple-vars/expected.d.ts
+++ b/tests/testcases/export-multiple-vars/expected.d.ts
@@ -1,0 +1,28 @@
+/**
+ * @public
+ */
+declare type In = { a: string };
+
+/**
+ * @public
+ */
+declare type Out = { b: number };
+/**
+ * @public
+ */
+declare const config: {
+  normalize: (inVar: In) => Out;
+};
+/**
+ * @public
+ */
+declare const options: {
+  normalize: (inVar: In) => Out;
+};
+/**
+ * @public
+ */
+declare const pararms: {
+  normalize: (inVar: undefined) => undefined;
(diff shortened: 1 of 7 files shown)


Swatinem:
Oh yes, there is a couple of things going on here.

Can you reduce the testcase to just the `.d.ts` files? The `.ts` source files are not used by the plugin (unless they are the *only* files).

I wonder why you have inline imports (`import("../settings").In`, both without file extension and with `.d.ts` extension) in your `util/index.d.ts`, while the corresponding `.ts` has straight up imports?
And how was it possible to infer the `Out` return type, while the `.ts` file just returns an object literal.

Those dynamic imports should work, there are tests for it, but its a big hack all the same, so maybe there is something wrong there.

Also, your `settings.d.ts` does not `export` anything. I know this is possible in ts under *some* circumstances, but this plugin only supports things that you explicitly `export` and `import`.

In either case, the plugin should handle this better and at least throw an error instead of generating broken code. I will take a look.

andykais:
No problem, Ill simplify & double check this tonight. What I did was copy the output from tsc and then remove unnecessary code & rename things for simplicity. I could have missed a few things in the process here though

Swatinem:
@andykais I recently fixed some issues with inline imports, can you give `1.1.10` another shot?
I also cleaned up your testcase a bit (added `export` declarations and fixed the filename issue): https://github.com/Swatinem/rollup-plugin-dts/commit/9683dca9c31aae5444acfa346484f90e4639c34e

Issue #49, opened 2019-07-25
import typeof issue


gavar:
## Test
```
export type TypeScript = typeof import ("typescript");
```

## Expect
```
declare type TypeScript = typeof import ("typescript");
export { TypeScript };
```

## Actual
```
import * as ɨmport1_typescript__ from 'typescript';
declare type TypeScript = typeof ɨmport1_typescript__;
export { TypeScript };
```

Swatinem:
There is already a testcase for this:

https://github.com/Swatinem/rollup-plugin-dts/blob/99cdf700d9cdb5d58010b9125264878668a2917d/tests/testcases/inline-import-external-namespace/expected.d.ts#L1-L6

I agree that the generated code is not really that beautiful, but as the already existing testcase shows, you can have more of these inline imports in one file, sometimes combined with the `typeof` modifier.

Rollup rewrites these if there is more than one, and you have to generate some kind of `identifier` for it. This was the best I came up with. And these type-aliases are perfectly fine and valid code.

gavar:
It seems the test case I submitted was not complex enough, here's another one

## Test
```
export type TypeScript = typeof import ("typescript");
export interface Test {
  rollup: import("rollup").RollupOptions;
}
```
## Expect
```
import * as ɨmport1_typescript__ from 'typescript';
import { RollupOptions } from 'rollup';
declare type TypeScript = typeof ɨmport1_typescript__;
interface Test {
  rollup: RollupOptions;
}
export { Test, TypeScript };
```

## Actual
```
import * as ɨmport1_typescript__ from 'typescript';
import { RollupOptions } from 'rollup';
declare type TypeScript = typeof ɨmport1_typescript__;
RollupOptions
export { Test, TypeScript };
```

> And these type-aliases are perfectly fine and valid code.

Intellij is going nuts reporting usage of non-ASCII symbols
Maybe it could be ASCII symbol like `$import` or `__import` for that?
![image](https://user-images.githubusercontent.com/3128402/61879995-b8fb9d80-aefc-11e9-84ae-1a7c1eb11ad4.png)

PR #50, opened 2019-07-25
test: type-import-typeof testcase

gavar:
test case for issue #49


Swatinem:
already covered here: https://github.com/Swatinem/rollup-plugin-dts/blob/99cdf700d9cdb5d58010b9125264878668a2917d/tests/testcases/inline-import-external-namespace/expected.d.ts#L1-L6

PR #52, opened 2019-07-30
Add support for bigints

cprecioso:
[Supported since TypeScript 3.2](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-2.html#bigint)

### DIFF
--- a/src/DeclarationScope.ts
+++ b/src/DeclarationScope.ts
@@ -28,6 +28,7 @@ const IGNORE_TYPENODES = new Set([
   ts.SyntaxKind.NeverKeyword,
   ts.SyntaxKind.ThisKeyword,
   ts.SyntaxKind.ThisType,
+  ts.SyntaxKind.BigIntKeyword,
 ]);
 
 interface DeclarationScopeOptions {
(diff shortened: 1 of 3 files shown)


Swatinem:
Thanks!

Can you add a testcase which also covers all the other ignored typenodes?

cprecioso:
I'm sorry, I don't have much experience writing tests, or know really what to test even. Could you help me with that?

Swatinem:
giving an example of the syntax that the plugin is choking on prior to this fix would be a good first start…

cprecioso:
just any reference to the bigint type https://gist.github.com/cprecioso/45da1379281e4ff67c28d6933c410d46

Swatinem:
got it, thanks!
I will add some tests later tonight when I get some time…

cprecioso:
I've mustered some very simple test case, can you please review it and see if it's sufficient?
https://github.com/Swatinem/rollup-plugin-dts/pull/52/commits/d4dd901f6f052910fbbdb12c653f8e721a0a912e

Swatinem:
That is perfect :-)

cprecioso:
If you could publish to npm soon, so it's able to be used, that'd be super! ☺️

Swatinem:
> If you could publish to npm soon, so it's able to be used, that'd be super! :relaxed: 

done

PR #55, opened 2019-08-14
Set up CI with Azure Pipelines

Swatinem:


Swatinem:
to summarize:
* having statistics about test failures etc on azure is nice, but jest junit is a bit of a pain -_-
* also, having builtin coverage on azure is a nice concept, but the generated reports are super ugly, and we have codecov anyway…

so all-in-all, github actions seem to be nicer, and hey, maybe they will add builtin test and coverage statistics at some point…

Issue #59, opened 2019-10-03
Spread operator in tuple types not supported


anion155:
HI! Thanks for the module.

But it's still does not supports spread in tuple type am I right? https://github.com/microsoft/TypeScript/pull/24897

roolup-plugin-dts: 1.1.8
typescript: 3.6.3
pnpm: 3.8.1

Error message:
```
[!] (plugin dts) Error: Could not load ./packages/lib-draw/build/utils/TypedEventEmitter.d.ts (imported by ./packages/lib-draw/build/Client.d.ts): Syntax not yet supported
1 | import { EventEmitter } from 'events';
> 2 | declare type Constructor<T = any, As extends [...any[]] = [...any[]], St = {}> = {
|                                               ^^^^^^^^
3 |     new (...args: As): T;
4 | } & Pick<St, keyof St>;
5 | declare type Arguments<T> = T extends (...args: any) => any ? Parameters<T> : T extends new (...args: any) => any ? ConstructorParameters<T> : any[];
Error: Could not load ./packages/lib-draw/build/utils/TypedEventEmitter.d.ts (imported by ./packages/lib-draw/build/Client.d.ts): Syntax not yet supported
1 | import { EventEmitter } from 'events';
> 2 | declare type Constructor<T = any, As extends [...any[]] = [...any[]], St = {}> = {
|                                               ^^^^^^^^
3 |     new (...args: As): T;
4 | } & Pick<St, keyof St>;
5 | declare type Arguments<T> = T extends (...args: any) => any ? Parameters<T> : T extends new (...args: any) => any ? ConstructorParameters<T> : any[];
at DeclarationScope.convertTypeNode (./node_modules/.registry.npmjs.org/rollup-plugin-dts/1.1.8_rollup@1.22.0+typescript@3.6.3/node_modules/rollup-plugin-dts/dist/rollup-plugin-dts.js:700:19)
at DeclarationScope.convertTypeNode (./node_modules/.registry.npmjs.org/rollup-plugin-dts/1.1.8_rollup@1.22.0+typescript@3.6.3/node_modules/rollup-plugin-dts/dist/rollup-plugin-dts.js:646:22)
at DeclarationScope.convertTypeParameters (./node_modules/.registry.npmjs.org/rollup-plugin-dts/1.1.8_rollup@1.22.0+typescript@3.6.3/node_modules/rollup-plugin-dts/dist/rollup-plugin-dts.js:617:18)
[...]

PR #60, opened 2019-10-03
fix(#59): add support for tuple type spread operator

anion155:
Fix #60

### DIFF
--- a/src/DeclarationScope.ts
+++ b/src/DeclarationScope.ts
@@ -332,6 +332,10 @@ export class DeclarationScope {
       this.pushReference(this.convertEntityName(node.exprName));
       return;
     }
+    if (ts.isTypeNode(node) && node.kind === ts.SyntaxKind.RestType) {
+      this.convertTypeNode((node as ts.RestTypeNode).type);
+      return;
+    }
     // istanbul ignore else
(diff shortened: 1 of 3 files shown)


src/DeclarationScope.ts
```diff
@@ -332,6 +332,10 @@ export class DeclarationScope {
       this.pushReference(this.convertEntityName(node.exprName));
       return;
     }
+    if (ts.isTypeNode(node) && node.kind === ts.SyntaxKind.RestType) {
+      this.convertTypeNode((node as ts.RestTypeNode).type);
```
Swatinem: Does TS have an explicit API for this, such as `ts.isRestTypeNode()`? That would be preferable to doing a manual cast here.

anion155: There is no any in [typescript.d.ts](https://github.com/microsoft/TypeScript/search?q=RestTypeNode&unscoped_q=RestTypeNode). So I'm assuming checking property `kind` is good enough.

Swatinem: Unfortunate. I filed this as an upstream feature request, since this really should be part of the upstream API: https://github.com/microsoft/TypeScript/issues/33785

Swatinem:
Thanks for the PR :-)

PR #68, opened 2019-11-06
Handle new getter accessor typings in TS 3.7

### LINKED ISSUE #67: Getters not supported?
```
[!] (plugin dts) Error: Could not load ~/dev/wana/src/auto.ts 
(imported by ~/dev/wana/src/core.ts): Syntax not yet supported
```

```
> 24 |     get test(): number | null;
     |     ^^^^^^^^^^^^^^^^^^^^^^^^^^
```

Repo here: https://github.com/alloc/wana

`rollup-plugin-dts` version: tested both `1.1.5` and `1.1.11`

`typescript` version: `3.7.2`

krawaller:
Fixes #67

### DIFF
--- a/src/DeclarationScope.ts
+++ b/src/DeclarationScope.ts
@@ -237,7 +237,8 @@ export class DeclarationScope {
         ts.isMethodSignature(node) ||
         ts.isConstructorDeclaration(node) ||
         ts.isConstructSignatureDeclaration(node) ||
-        ts.isCallSignatureDeclaration(node)
+        ts.isCallSignatureDeclaration(node) ||
+        ts.isGetAccessorDeclaration(node)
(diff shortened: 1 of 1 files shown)


Swatinem:
thanks! I’m currently working on some other stuff as well, will add some tests and release later on today…

krawaller:
Humble apologies @Swatinem, but I think I provided a bogus fix. This will type the getter as a method, but in reality it should be typed as a read-only property, right? :/

Swatinem:
I doesn’t matter, as long as all the arguments and return types are correctly handled.

krawaller:
Doh, right, which they will be. Main thing was just to not short circuit. I see now, thank you! :)

PR #74, opened 2020-01-19
Add including discovered d.ts files in ts compile mode

### LINKED ISSUE #71: Unable to get rollup-plugin-dts to reference local d.ts file
Hello! 👋 I'll preface this with that it may be entirely my fault this isn't working, but I've been trying to sort it out for a bit.

I have a few `npm` installed imports in a `.ts` file I've written that *do not* have types available (either locally or via DefinitelyTyped), so I've done what I normally do and have written a local `.d.ts` file to tell TypeScript to not worry about it. (This import doesn't appear in the types at all, so there's technically nothing for `rollup-plugin-dts` to "render" based on it.) My `tsconfig.json` has `strict: true` set, so it will throw an error otherwise about implicit `any`. Something like this in a file called `vendor.d.ts` that sits next to the input file:

```
declare module 'lib-without-types';
```

In VSCode and with `tsc --noEmit` this is enough — my code passes the check. But when I try to run `rollup-plugin-dts` against my `.ts` file, It throws `TS7016` about not being able to find a declaration file. As far as I can tell `rollup-plugin-dts` only considers the input script and that alone, even when the `tsconfig.json` has an `includes` value set (which should "tell" TypeScript to include my local `d.ts` file when resolving types).

(Also **super** weird — `rollup-plugin-dts` didn't care/complain about this until TypeScript 3.7 landed. If I roll back to TypeScript 3.6 it doesn't throw `TS7016`, even if I **don't** have the `vendor.d.ts` file at all!)

Is there something I'm missing or is that by design? Thank you!

rdmurphy:
Fixes #71 

I'm not sure exactly how to test this. 🤔 The primary way I was able to trigger the type check error (`noImplicitAny`) was by having a third-party library without a packaged `d.ts` file that I wanted to account for with a local declaration. All current tests continue to pass however with these changes.

I added a `Set` to track the found `d.ts` files but that may be overkill — just seemed like the easiest way to prevent duplication in modification to `inputs`. Not sure if TypeScript would have cared or not.

### DIFF
--- a/src/program.ts
+++ b/src/program.ts
@@ -26,24 +26,30 @@ const OPTIONS_OVERRIDE: ts.CompilerOptions = {
   target: ts.ScriptTarget.ESNext,
 };
 
-function getCompilerOptions(input: string): { dirName: string; compilerOptions: ts.CompilerOptions } {
+function getCompilerOptions(
+  input: string,
+): { dtsFiles: Array<string>; dirName: string; compilerOptions: ts.CompilerOptions } {
   let dirName = path.dirname(input);
+  let dtsFiles: Array<string> = [];
   const configPath = ts.findConfigFile(path.dirname(input), ts.sys.fileExists);
   if (!configPath) {
-    return { dirName, compilerOptions: { ...OPTIONS_OVERRIDE } };
+    return { dtsFiles, dirName, compilerOptions: { ...OPTIONS_OVERRIDE } };
   }
   dirName = path.dirname(configPath);
   const { config, error } = ts.readConfigFile(configPath, ts.sys.readFile);
   if (error) {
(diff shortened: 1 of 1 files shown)


src/program.ts
```diff
@@ -65,6 +72,7 @@ export function createPrograms(input: Array<string>) {
 
     main = path.resolve(main);
     const options = getCompilerOptions(main);
+    options.dtsFiles.forEach(dtsFiles.add, dtsFiles);
```
rdmurphy: If this is too much "magic" happy to swap it out for a `for...of` loop or a more standard `Array.forEach`.

Swatinem:
Sorry for the delay. I just merged as-is and will release later. I didn’t have any time to properly look into this.

PR #75, opened 2020-01-25
Support string literal class properties

FredKSchott:
Before this change, this property in lit-element would fail when seen by rollup-plugin-dts:

```
export class LitElement extends UpdatingElement {
  protected static['finalized'] = true;
```

This is valid TS that fails with this plugin.

### DIFF
--- a/src/DeclarationScope.ts
+++ b/src/DeclarationScope.ts
@@ -194,7 +194,7 @@ export class DeclarationScope {
     }
     const { expression } = node.name;
 
-    if (ts.isIdentifier(expression)) {
+    if (ts.isIdentifier(expression) || ts.isStringLiteral(expression)) {
       return this.pushReference(createIdentifier(expression));
     }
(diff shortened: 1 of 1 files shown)


src/DeclarationScope.ts
```diff
@@ -194,7 +194,7 @@ export class DeclarationScope {
     }
     const { expression } = node.name;
 
-    if (ts.isIdentifier(expression)) {
+    if (ts.isIdentifier(expression) || ts.isStringLiteral(expression)) {
```
Swatinem: Actually we rather want to skip this, instead of running it through the branch. By definition, there is no way a string literal will refer to anything you have imported.

Also, it is probably the same for numbers and other literals as well? Some tests to cover this would be nice as well.

Swatinem:
pulled and rebased here: https://github.com/Swatinem/rollup-plugin-dts/commit/a9746802c4af42c5f74dfeab4513f0ca81810c4f

Thanks!

Issue #44, opened 2019-06-22
Expose extended modules in definition bundle


doctorcode-co:
Hi
Consider this code:
```
declare module react {
    export interface test {
    }
}
```
how i can expose this in final bundle?

Swatinem:
I would have to look up how these TS module declarations work.

Because you don’t really `export` anything, it just modifies a global object as a side-effect, which is super weird…

doctorcode-co:
In source of my plugin i extends react or knockout or vue and etc. this is declared in a ts file named react.extensions.ts (or ko.extenstions.ts) then in any file we need this plugin import it.
source of react.extensions.ts:
```
declare module react {
    export interface test {
    }
}
```
source of plugin:
```
import { test } from 'react';
...
const t = new test();
```
if **import { test } from 'react';** not declared then can not use test. Also, the user does not need to enter the separate definition file. 
The tsc emit it for me but dts plugin not emit this file in bundle.

doctorcode-co:
dear @Swatinem :
Note that for extend jsx, need to emit declared global modules in final bundle.
```
declare global {
    namespace JSX {
        interface Element extends HTMLElement { }
        interface ElementClass extends ComponentBase {
        }
        interface ElementAttributesProperty { props: {}; }
    }
}
```

Swatinem:
I played around with this a little bit, its a hard problem to solve though.

As a workaround, you can put all of your global augmentations into a separate file, and make it `external`, so it does not run through this plugin at all.
I think that should work.

doctorcode-co:
your are true, it's hard to solved. I use gulp and fs to read source file and add to final destination file, temporary solved my problem. 
Thank you

morlay:
```ts
declare global {
  namespace NodeJS {
    interface Global {
      second: Second;
    }
  }
}

export interface Second {}
```

```ts
// index.d.ts
export * from "./second";

export interface First {}

declare global {
  namespace NodeJS {
    interface Global {
      first: First;
    }
  }
}
```

could be transform to  (declare global will be merge by typescript, so i think coping could be fine)

```ts
declare global {
  namespace NodeJS {
    interface Global {
      first: First;
    }
  }
}
declare global {
  namespace NodeJS {
    interface Global {
      second: Second;
    }
  }
}

interface Second {}
interface First {}
export { First, Second };
```

but have to export `First`, `Second`.

@doctorcode-co @Swatinem 
how do think this behavier，just copy the global or module declaration?

PR #81, opened 2020-03-11
feat: module declaration or global declaration should 1:1 copy

### LINKED ISSUE #72: module declarations with label that is not an ID
The following declaration content fails:
```javascript
declare module 'd3-selection' {
...
}
```
With the following error:
```
[!] (plugin dts) Error: Could not load C:\Users\gordon\git\hpcc-js\node_modules\@types\d3-transition\index.d.ts (imported by C:\Users\gordon\git\hpcc-js\packages\graph\types\Graph.d.ts): namespace name should be an "Identifier"
```

### LINKED ISSUE #80: [feature] keep module declaration with jsdoc @override
Some time we may extends prototype of deps like below:

```ts
declare module "rxjs/internal/Observable" {
  interface Observable<T> {
     custom()
  }
}
```
the module declaration should be keep. (now tsc will keep them, but rollup-plugin-dts dropped).

morlay:
fix #80 #44 
may fix #72

### DIFF
--- a/src/Transformer.ts
+++ b/src/Transformer.ts
@@ -6,6 +6,7 @@ import {
   createExport,
   createIdentifier,
   createProgram,
+  isNamespaceDeclaration,
   matchesModifier,
   Ranged,
   withStartEnd,
@@ -27,12 +28,14 @@ export interface TransformOutput {
   ast: ESTree.Program;
   fixups: Array<Fixup>;
   typeReferences: Set<string>;
+  moduleDeclarations: Set<string>;
 }
 
 export class Transformer {
   ast: ESTree.Program;
   fixups: Array<Fixup> = [];
   typeReferences = new Set<string>();
+  moduleDeclarations = new Set<string>();
 
   declarations = new Map<string, DeclarationScope>();
   exports = new Set<string>();
@@ -65,6 +68,7 @@ export class Transformer {
       ast: this.ast,
       fixups: this.fixups,
       typeReferences: this.typeReferences,
+      moduleDeclarations: this.moduleDeclarations,
     };
   }
 
@@ -83,6 +87,7 @@ export class Transformer {
   unshiftStatement(node: ESTree.Statement | ESTree.ModuleDeclaration) {
     this.ast.body.unshift(withStartEnd(node, { start: 0, end: 0 }));
   }
+
   pushStatement(node: ESTree.Statement | ESTree.ModuleDeclaration) {
     this.ast.body.push(node);
   }
@@ -149,7 +154,13 @@ export class Transformer {
       return this.convertExportDeclaration(node);
     }
     if (ts.isModuleDeclaration(node)) {
-      return this.convertNamespaceDeclaration(node);
+      if (isNamespaceDeclaration(node)) {
+        return this.convertNamespaceDeclaration(node);
+      }
+
+      this.moduleDeclarations.add(node.getFullText());
+      this.removeStatement(node)
+      return;
     }
     if (node.kind == ts.SyntaxKind.NamespaceExportDeclaration) {
       // just ignore `export as namespace FOO` statements…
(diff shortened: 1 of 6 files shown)


src/index.ts
```diff
@@ -182,15 +188,20 @@ const plugin: PluginImpl<Options> = (options = {}) => {
       const source = ts.createSourceFile(chunk.fileName, code, ts.ScriptTarget.Latest, true);
       const fixer = new NamespaceFixer(source);
 
-      code = Array.from(typeReferences, ref => `/// <reference types="${ref}" />`).join("\n");
-      if (code) {
-        code += "\n";
-      }
+      code = writeBlock(Array.from(typeReferences, ref => `/// <reference types="${ref}" />`));
+      code += writeBlock(Array.from(moduleDeclarations, moduleDeclaration => moduleDeclaration));
```
```suggestion
      code += writeBlock(Array.from(moduleDeclarations));
```
I don’t think you need an identity transform here.

src/Transformer.ts
```diff
@@ -149,7 +154,13 @@ export class Transformer {
       return this.convertExportDeclaration(node);
     }
     if (ts.isModuleDeclaration(node)) {
-      return this.convertNamespaceDeclaration(node);
+      if (isNamespaceDeclaration(node)) {
+        return this.convertNamespaceDeclaration(node);
```
Swatinem: we would need to at least walk all the children of module declarations, so we correctly pull in any referenced items.

src/astHelpers.ts
```diff
@@ -218,3 +218,7 @@ export function matchesModifier(node: ts.Node, flags: ts.ModifierFlags) {
   const nodeFlags = ts.getCombinedModifierFlags(node as any);
   return (nodeFlags & flags) === flags;
 }
+
+export const isNamespaceDeclaration = (node: ts.ModuleDeclaration) => {
+  return node.modifiers && (node.name.pos - node.modifiers[node.modifiers?.length - 1].end) === 10 /* "namespace".length + 1 */
```
Swatinem: I really don’t like this magic number counting, for example it is not resilient to whitespace and comments in between the modifiers, and `node.name`. I’m sure there is a better way to handle this somehow.

src/index.ts
```diff
@@ -182,15 +188,20 @@ const plugin: PluginImpl<Options> = (options = {}) => {
...
       code += fixer.fix();
 
       return { code, map: { mappings: "" } };
     },
   };
 };
 
+function writeBlock(codes: string[]): string {
```
```suggestion
function writeBlock(codes: Array<string>): string {
```
just personal preference ;-)

tests/testcases/decl-module-or-global/index.ts
```diff
@@ -0,0 +1,10 @@
+declare module "babel__core" {
+  var fn1: any
```
Swatinem: Like I said above, it might be good to actually reference a local type here, so that we make sure that it is correctly pulled in.
For example `const export1: A;` here, and somewhere else just define an empty `interface A {}`

Swatinem:
@morlay don’t worry about updating this, I will take care of it

morlay:
@Swatinem cool. I check your updates, that will be clear.

but some thing i don't understand.
here, should we collect `global` and namespace as `namespace`?

https://github.com/Swatinem/rollup-plugin-dts/blob/master/src/NamespaceFixer.ts#L48
```ts
if (ts.isModuleDeclaration(node) && ts.isIdentifier(node.name)) {
    itemTypes[node.name.getText()] = "namespace";
}
```

if not, for strictly namespace, it could be (as Transformer did https://github.com/Swatinem/rollup-plugin-dts/blob/master/src/Transformer.ts#L184)
```ts
if (ts.isModuleDeclaration(node) && (node.flags & ts.NodeFlags.Namespace)) {
    itemTypes[node.name.getText()] = "namespace";
}
```

PR #84, opened 2020-04-07
Add TS compiler options to plugin options

martinadamec:
Hi, for my project I need set up `baseUrl` for TypeScript, so I added `compilerOptions` to handle it. 

I'm using absolute path, for example I have file `src/components/Button.tsx` and elsewhere I use import `import Button from 'components/Button'`.

If you find out it as valuable for you, plese look at the PR ;)
If I did something wrong, let me know, thank you (and for your work).

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -16,10 +16,11 @@ export interface Options {
    * configured via rollup.
    */
   respectExternal?: boolean;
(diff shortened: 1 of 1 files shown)


Swatinem:
That sure is a good usecase.
I wonder how this interacts with the createPrograms that we do. Also it would be nice to have a test that covers this.

PR #116, opened 2020-09-22
Create a new preproces step

Swatinem:
This makes a few things a lot nicer as it preprocesses the code a bit

fixes #100
fixes #112

PR #125, opened 2020-10-29
test: respectExternal with `import type`

aleclarson:
The test in this PR fails with `Unexpected token` because of the `import type` declaration in the `bar/index.ts` module.

It does *not* fail when `import type` is used in the `index.ts` module.

This test uses `respectExternal: true` and `external: []`.

It seems like TypeScript parsing is not enabled for non-external bare imports?


aleclarson:
Here's where it's tripping up:
https://github.com/microsoft/TypeScript/blob/3517af8f80460ab28c49eeccddaa95db5654c730/src/compiler/utilities.ts#L4176

The `writeFileCallback` passed to `program.emit` is never called for `node_modules` paths, which breaks the `respectExternals` option. I guess we need to create a `Program` for each non-external dependency.

Swatinem:
Just curious, what is your usecase for having `.ts` files in `node_modules`?

aleclarson:
I'm bundling a TypeScript monorepo

Swatinem:
In that case, isn’t it faster to use project references and incremental builds? Which would by definition create incremental `.d.ts` files that you can consume instead of relying on translating at load time?

aleclarson:
I've never used project references, so I'm not sure. Are you saying I shouldn't be using this plugin at all for that use case?

aleclarson:
Okay, I tried using the compiled output of each dependency in the monorepo, but this plugin is now choking on the `.d.ts` syntax. It seems to be expecting JavaScript. I'll update this test case.

Swatinem:
> you're saying I should build each package in the monorepo separately first, then use `rollup-plugin-dts` on the output.

yes. compiling `.ts` -> `.d.ts` internally is a huge pain to maintain, and consuming pre-generated `.d.ts` files is a lot simpler.

aleclarson:
Two downsides to compiling each package before bundling:
- Having to ensure each package has been compiled since last change
- Generating semi-broken sourcemaps, where the `sources` array points to compiled modules instead of the original source

For these reasons, I think it's worth supporting this test case.

aleclarson:
Hmm, when I remove `sourcemapExcludeSources: false` from my Rollup config, the sourcemap resolves the `sources` array using the sourcemaps of the compiled dependencies, so I think I can live without support for this test case.

Swatinem:
BTW, I haven’t really focused on sourcemaps *at all* so far, but I started down the road simplifying some of the internals in #116 which would make sourcemaps a bit better, at least if I also combine that with a proper post-process step.

Regarding your concerns about compiling each module. Thats actually specifically the usecase for TS project references. The problem is just that they are horrible and sometimes even hurt performance, and sometimes generate wrong import statements, especially if you use path mapping. We had quite some problems with those back when I tried them. Anyway, I hope they will get better / more usable with time.

PR #126, opened 2020-10-29
test: link to externals outside node_modules 

aleclarson:
In this test, the `node_modules/d3-drag` path is linked to `../../d3-drag` (which is a local package, not from node_modules).

This setup mimics a monorepo, where dependencies often point to local packages.

The local `d3-drag` package contains an `index.d.ts` module.

This test fails with `Unexpected token` error, since no `Program` is found for the `d3-drag/index.d.ts` module.


aleclarson:
I'm working on a fix, so lmk if you are too. ;)

aleclarson:
Okay, I've got a fix working! LGTM 👍

src/index.ts
```diff
@@ -53,6 +46,14 @@ const plugin: PluginImpl<Options> = (options = {}) => {
           ts.ScriptTarget.Latest,
           true, // setParentNodes
         );
+    } else {
+      // Rollup doesn't tell you the entry point of each module in the bundle,
+      // so we need to ask every TypeScript program for the given filename.
+      program = programs.find(p => (source = p.getSourceFile(fileName)));
+      if (!program && fs.existsSync(fileName)) {
```
Swatinem, line 53:
```suggestion
      if (!program && ts.sys.fileExists(fileName)) {
```

avoids the import, since we already use `ts.sys` in this file.

tests/testcases/externals-link/meta.ts
```diff
@@ -0,0 +1,6 @@
+exports.options = {
```
Swatinem, line 1: I would prefer JSON for these files, as in most of the other cases as well.

tests/testcases/externals-link/node_modules/d3-drag
```diff
@@ -0,0 +1 @@
+../../../d3-drag
```
Swatinem, line 1: any specific reason for this link? I would rather like to keep the testcase self contained.

Swatinem (commented):
Nice!

tests/testcases/externals-link/node_modules/d3-drag
Swatinem: lol nevermind, I should have read the PR description, lol.

aleclarson: unfortunately, it has to exist outside the root directory of the test case for the error to occur :(

Issue #137, opened 2021-03-05
Feature request: Allow passing in a readFile function as an option


marijnh:
I have a setup where I'm using this plugin as part of a build pipeline that doesn't write its intermediate files to disk. I'm currently doing a dodgy thing where I temporarily overwrite `ts.sys.readFile` to hook in access to the files I have in memory, and it works, but I'd like to not do that. I'll gladly prepare a PR if this is something you'd be open to.

Swatinem:
I think that is a sensible thing to do, yes.

I wanted to revamp the loading/transforming parts of this plugin at some point either way, but I don’t think I will have time for that anytime soon.

So this would override the `readFile` hook that is being passed to the `createProgram` calls as well, right?

marijnh:
Yes, that'd be the idea.

Another possible approach, which would be cleaner and also serve my purpose, would be to replace the plugin's `load` hook with a `transform` hook... but I'm guessing there is a reason it's a `load` hook?

Swatinem:
I have to double-check, this was all quite long ago, but back then I think I had to use the `load` hook to avoid the code being parsed as JS.

marijnh:
The [docs](https://rollupjs.org/guide/en/#transform) suggest it runs before parsing (which makes sense because most types of transformation plugins will need to act before parsing).

marijnh:
(If you don't have time, but are open to the change, I can take a stab at moving this to a transform hook, and let you know if it's possible.)

Swatinem:
Sure, that would be awesome!

BTW, my reasoning why I was looking into having both load+transform hooks was to better support sourcemaps:
* `load` the `.d.ts{,.map}`
* `transform` and return another `map` via `magic-string` which I started using in the preprocessor.
* (eventually use `magic-string` again in the post-processor to have sourcemaps front to back)

marijnh:
Thanks a lot! If you have the time to cut a release in the near future, I'll glad to be able to drop my hack.

Swatinem:
> If you have the time to cut a release in the near future, I'll glad to be able to drop my hack.

Published `3.0.0`; major bump since I used the opportunity to switch to ES modules and drop Node 10 support.

marijnh:
Appreciated.

PR #138, opened 2021-03-05
Use transform hook instead of load hook for code transformation

marijnh:
See discussion in #137. This was really easy. But I'm a bit confused by the use of magic-string in the package—it doesn't seem to actually _do_ anything with the source maps, it just converts the magic string object back to a string and discards the mapping data (if I'm reading it right). Is that code part of a planned future extension, or am I missing something?

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -31,20 +31,18 @@ const plugin: PluginImpl<Options> = (options = {}) => {
   // except when all entry points are ".d.ts" modules.
   let programs: Array<ts.Program> = [];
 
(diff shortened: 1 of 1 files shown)


Swatinem:
> Is that code part of a planned future extension, or am I missing something?

Yes. SourceMaps are pretty much broken now since the postprocessing can potentially shuffle code around however it likes, so I didn’t bother trying to make it work at all so far.

Swatinem (approved):

Swatinem:
Nice! That was indeed easier than expected

PR #164, opened 2021-08-02
Support namespaces without a ModuleBlock

### LINKED ISSUE #160: global namespace is not work
### Checklist

* [ ] This issue is not related to rolling up `@types`.

### Code Snipped

```ts
import * as Types from './types'
import * as Models from './models'

declare global {
  namespace Named.Core {
    export { Types, Models }
  }
}

```

### Error Message

```
// The Error Message thrown by rollup / the plugin.
[!] (plugin dts) Error: namespace must have a "ModuleBlock" body.
```

Swatinem:
Might fix #160

### DIFF
--- a/src/transform/DeclarationScope.ts
+++ b/src/transform/DeclarationScope.ts
@@ -315,9 +315,13 @@ export class DeclarationScope {
     }
   }
 
-  convertNamespace(node: ts.ModuleDeclaration) {
+  convertNamespace(node: ts.ModuleDeclaration, relaxedModuleBlock = false) {
     this.pushScope();
 
+    if (relaxedModuleBlock && node.body && ts.isModuleDeclaration(node.body)) {
+      this.convertNamespace(node.body, true);
+      return;
+    }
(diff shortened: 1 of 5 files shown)


Mattstir500:
Hi, is there already a plan when this will get merged and released? Thanks! :)

NijatInHIT:
@Swatinem 
Hi, I think I still have this problem, 
Please have a look at the error

```
./src/types/global.d.ts → lib/index.d.ts...
[!] (plugin dts) Error: namespace must have a "ModuleBlock" body.

  2 | type inputTS = 'orderPrice' | 'orderQty'
  3 |
> 4 | declare module '@option/design'
    | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

this is my dts config
```
const dtsConfig = {
  input: './src/types/global.d.ts',
  output: [{ file: 'lib/index.d.ts', format: 'es' }],
  plugins: [dts(), nodeResolve(), commonjs(), typescript()],
}
```

PR #186, opened 2021-11-15
issue 185 - added test case

funnisimo:
Added test of sample provided in issue.


Swatinem (approved):

Swatinem:
Test looks good! I have an idea of what needs to be done to fix this, just need to find some time to actually work on this.

funnisimo:
Is there any way I can help?

I also noticed (but did not add to the issue/test) that the following also causes hoist problems:

>> a.ts
export type Callback<T extends Props> = (obj: T) => any;

PR #188, opened 2021-12-16
Support for reference path directives

lmartorella:
PR to fix the [triple-slash reference path](https://www.typescriptlang.org/docs/handbook/triple-slash-directives.html#-reference-path-), not correctly bundled in the output file.
For example:

`index.d.ts`:
```ts
/// <reference path="./lib-global-ns.d.ts" />
export * from "./sub/lib-es.ts";
```

`lib-es.ts`:
```ts
/// <reference path="./lib-global-ns-2.d.ts" />

// [...]
```
should result in a bundled file with all the references at the top, with correct relative paths:

```ts
/// <reference path="./lib-global-ns.d.ts" />
/// <reference path="./sub/lib-global-ns-2.d.ts" />

// [...]
```

The triple-slash reference path syntax can still be used in ES-module declarations files. For example, it can be used in mixed projects that are still using global namespaces. In that case, migrating a sub-section of the project should leave the structure of the reference paths untouched.

> Note: I'm sure how to get the target file path from the `renderChunk` function. During the UT the `options.file` seems to not be there, so I used the `chunk.facadeModuleId` in that case.

Thanks, L

### DIFF
--- a/src/transform/index.ts
+++ b/src/transform/index.ts
@@ -3,6 +3,7 @@ import ts from "typescript";
 import { NamespaceFixer } from "./NamespaceFixer.js";
 import { preProcess } from "./preprocess.js";
 import { convert } from "./Transformer.js";
+import * as path from "path";
 
 export interface TransformOptions {}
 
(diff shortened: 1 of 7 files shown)


Swatinem (approved):

Swatinem:
Ah very nice!
I thought to ask about if it wouldn’t be better to just inline the contents of these files, since you would have to ship them if they are referenced. But your comment explained it very well that this should only be a stopgap solution on the migration path to proper ES modules ;-)

lmartorella:
Oh yes, there are use cases in which different top-level modules of a bigger project has to refer the same global ".d.ts" (still using namespaces). Inlining it would "break" the inheritance. Thanks!

PR #196, opened 2022-02-11
feat: add .d.cts, .d.mts support

xiaoxiangmoe:


tests/testcases/multiple-entries-cjs/expected.d.ts
```diff
@@ -0,0 +1,8 @@
+// main-a.d.cts
+export { A } from './common.d-df0efa22.js';
```
Swatinem, line 2: I’m surprised why this is not `.cjs` then?

Swatinem (commented):
I remember some older typescript versions would actually treat it as an error when you provided a file extension. I also know at some point they relaxed that restriction and you now *can* put extensions, though I was not aware that you now have them for `.d.ts` files as well, and that the definition files actually point to the non-definition JS files.
Do you have a link to the docs that explain this?

tests/testcases/multiple-entries-cjs/expected.d.ts
xiaoxiangmoe: because default value of chunkFileNames is '[name]-[hash].d.ts'

src/transform/NamespaceFixer.ts
```diff
@@ -209,6 +218,7 @@ export class NamespaceFixer {
         code += `}`;
       }
 
+      code += ns.textBeforeCodeAfter ?? "";
       code += codeAfter;
```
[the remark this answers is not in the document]
xiaoxiangmoe: 😂 Do you have any idea about better naming?

xiaoxiangmoe:
> I remember some older typescript versions would actually treat it as an error when you provided a file extension. 

I test it in TypeScript 2.0.0. And it works well.

https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-0.html#module-identifiers-allow-for-js-extension

tests/testcases/multiple-entries-cjs/meta.js
```diff
@@ -0,0 +1,6 @@
+export default {
+  rollupOptions: {
+    input: ["main-a.d.ts", "main-b.d.ts"],
+    output: { entryFileNames: "[name].d.cts" },
```
xiaoxiangmoe, line 4: here only has entryFileNames

tests/testcases/multiple-entries-mjs/meta.js
```diff
@@ -0,0 +1,9 @@
+export default {
+  rollupOptions: {
+    input: ["main-a.d.ts", "main-b.d.ts"],
+    output: {
+      entryFileNames: "[name].d.mts",
+      chunkFileNames: '[name]-[hash].d.mts'
```
xiaoxiangmoe, line 6: here has chunkFileNames

tests/testcases/multiple-entries-mjs/main-a.d.ts
```diff
@@ -0,0 +1 @@
+export { A } from "./common";
```
Swatinem, line 1:
```suggestion
export { A } from "./common.mjs";
```

I wonder what happens when you also include the extension in the source here.

Swatinem (approved):
looks nice, lets worry about that naming another time :-D

But I am curious to see how this reacts to having extensions in the sources.

tests/testcases/multiple-entries-mjs/main-a.d.ts
Swatinem: Ah okay, lets move on with this solution then.

xiaoxiangmoe: I try `export { A } from "./common.js";` and it all works well in both rollup-plugin-typescript2 and rollup-plugin-dts.

xiaoxiangmoe:
@Swatinem Could you tell me when will this feature release? Should I publish a private version in our company's own npm registry?

Swatinem:
> Should I publish a private version in our company's own npm registry?

yes, that might be a good idea for now. I think I will hold off from releasing for now, I would like to maybe test-drive this myself and see how that works.
I also want to see what `tsc` actually does if I switch to `.mjs` extensions in the own codebase, though I’m not entirely sure that is well supported yet.

PR #203, opened 2022-03-20
Reference remapping should not touch absolute path

yurynix:
Hi 👋 
First of all thank you for the great package 🙃 , while using it I've noticed triple slash references are remapped to the main file, which is great.

However, while re-mapping references, IMO the remapping should handle _only_ relative paths, and leave as-is absolute paths.
Another option is to respect [noResolve](https://www.typescriptlang.org/tsconfig#noResolve) TS `compilerOptions` flag maybe?


Swatinem (approved):

yurynix:
Thank you for the swift response @Swatinem 🙏

PR #212, opened 2022-05-29
fix: export both cjs and esm types

### LINKED ISSUE #210: `"type": "module"` is preventing this package from being import from non-module environments
I've opened an [issue](https://github.com/microsoft/TypeScript/issues/49299) for this with TypeScript as I believe this to be unexpected behavior, but if it ends up being the case that it's not then `"type": "module"` will need to be removed from `package.json`.

RebeccaStevens:
Fixes #210

https://github.com/microsoft/TypeScript/issues/49299#issuecomment-1140421993#212

### DIFF
--- a/rollup.config.ts
+++ b/rollup.config.ts
@@ -15,7 +15,10 @@ const config: Array<RollupWatchOptions> = [
   },
   {
     input: "./.build/src/index.d.ts",
-    output: [{ file: pkg.types, format: "es" }],
+    output: [
+      { file: pkg.exports.types.import, format: "es" },
+      { file: pkg.exports.types.require, format: "commonjs", exports: "named" },
+    ],
     plugins: [dts()],
   },
 ];
--- a/package.json
+++ b/package.json
@@ -27,11 +27,14 @@
   "type": "module",
   "main": "./dist/rollup-plugin-dts.cjs",
   "exports": {
-    "types": "./dist/rollup-plugin-dts.d.ts",
+    "types": {
+      "import": "./dist/rollup-plugin-dts.d.mts",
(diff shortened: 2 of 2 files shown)


rollup.config.ts
```diff
@@ -15,7 +15,10 @@ const config: Array<RollupWatchOptions> = [
   },
   {
     input: "./.build/src/index.d.ts",
-    output: [{ file: pkg.types, format: "es" }],
+    output: [
+      { file: pkg.exports.types.import, format: "es" },
+      { file: pkg.exports.types.require, format: "commonjs", exports: "named" },
```
Swatinem, line 20: this won’t really have any effect, as we override these output options; and exclusively only generate ES exports.

RebeccaStevens: We need to generate both a `.d.mts` and a `.d.cts` file. These two files would be identical but they need to be different files.

I'll leave it to you to figure what the best way to do that is. Whether it be to use this PR as is, use `format: "es"` for both outputs or just generate one then copy-paste it.

RebeccaStevens: Could even just shorten it to:

```ts
output: [{ file: pkg.exports.types.import }, { file: pkg.exports.types.require }]
```

Swatinem: I think the rollup typings require you to define at least one `format`. I can take a look on the weekend again, I was distracted with other matters these days.

Finesse:
I experience this problem too. Do you need help with the PR?

Swatinem:
https://github.com/Swatinem/rollup-plugin-dts/pull/271 did just this slightly differently

PR #227, opened 2022-10-13
feat!: support rollup v3

### LINKED ISSUE #226: Rollup v3 support
https://github.com/rollup/rollup/tree/v3.0.0

```
├─┬ rollup-plugin-dts 4.2.3
│ └── ✕ unmet peer rollup@^2.55: found 3.0.0
```

```
packages/shared build: ./src/node/index.ts → ./lib/node/index.d.ts...
packages/shared build: (!) The "output.namespaceToStringTag" option is deprecated. Use the "output.generatedCode.symbols" option instead.
packages/shared build: [!] RollupError: Invalid value false for option "output.interop" - use one of "compat", "auto", "esModule", "default", "defaultOnly".
packages/shared build: https://rollupjs.org/guide/en/#outputinterop
packages/shared build: RollupError: Invalid value false for option "output.interop" - use one of "compat", "auto", "esModule", "default", "defaultOnly".
```

pi0:
Resolves #226

- Update transform outputOptions for v3 compatibility 
- Update rollup peerDependency to `3.0.0` (options API is changed)
- Add json assertion to `package.json` import

### DIFF
--- a/src/transform/index.ts
+++ b/src/transform/index.ts
@@ -67,8 +67,8 @@ export const transform: PluginImpl<TransformOptions> = () => {
         exports: "named",
         compact: false,
         freeze: true,
-        interop: false,
-        namespaceToStringTag: false,
+        interop: 'esModule',
+        generatedCode: Object.assign({ symbols: false }, options.generatedCode),
         strict: false,
       };
     },
--- a/package-lock.json
+++ b/package-lock.json
@@ -21,7 +21,7 @@
         "c8": "^7.12.0",
         "fs-extra": "^10.1.0",
         "rimraf": "^3.0.2",
-        "rollup": "2.79.1",
+        "rollup": "3.1.0",
         "typescript": "4.8.4"
       },
       "engines": {
@@ -34,7 +34,7 @@
         "@babel/code-frame": "^7.18.6"
       },
       "peerDependencies": {
-        "rollup": "^2.55",
+        "rollup": "^3.0.0",
         "typescript": "^4.1"
       }
     },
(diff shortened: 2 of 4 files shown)


src/transform/index.ts
```diff
@@ -67,8 +67,8 @@ export const transform: PluginImpl<TransformOptions> = () => {
         exports: "named",
         compact: false,
         freeze: true,
-        interop: false,
-        namespaceToStringTag: false,
+        interop: 'esModule',
+        generatedCode: Object.assign({ symbols: false }, options.generatedCode),
```
Swatinem, line 71: I believe you can use object spread for this.

Swatinem (approved):

Swatinem:
CI is not too happy. It needs to bump the lower node version. Also there is a bunch of snapshot changes, but those seem to be related to some more comments in the output. I can check all that stuff maybe on the weekend.

Mister-Hope:
Btw, can we add tests for 14 and 16? That may perform a more accurate result

SukkaW:
> CI is not too happy. It needs to bump the lower node version.

Rollup has bumped the minimum required Node.js version to 14 as part of its breaking changes.

If `rollup-plugin-dts` is going to adopt this change, a new major version has to be released.

src/transform/index.ts
pi0: Indeed, it was rather an annoying typescript issue i had to use alternative syntax since it nags about not spreading possibly undefined `generatedCode` (while technically `{ ...undefined}` works in the wonderful land of js!). Happy to do another method if prefer.

Mister-Hope: `{ symbols: false, ...(options.generatedCode || {})` is not working?

SukkaW: IMHO, if I'd to choose, I'll prefer `Object.assign({ symbols: false }, options.generatedCode) ` over `{ symbols: false, ...(options.generatedCode || {})}`.

Swatinem:
I manually merged this, fixed up all the tests, and bumped the minimum supported rollup and node versions.

> can we add tests for 14 and 16?

I rather just want to test with latest node and minimum supported, I don’t think testing intermediate versions provides any value.

I will publish a new major in the coming days.

PR #233, opened 2022-12-01
test: run with tsx

privatenumber:
If you would like, I updated the test command to run with [tsx](https://github.com/esbuild-kit/tsx) so it doesn't require a compilation step.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -30,5 +30,4 @@ jobs:
         run: bash <(curl -s https://codecov.io/bash) -t ${{secrets.CODECOV_TOKEN}} -B ${{ github.ref }} -f coverage/coverage-final.json
       # test the minimum supported peer dependency version
       - run: npm install typescript@4.1 rollup@3.0
-      # aka `npm test` without the `pretest/build`
-      - run: node .build/tests/index.js
+      - run: npm test
--- a/package-lock.json
+++ b/package-lock.json
@@ -22,6 +22,7 @@
         "fs-extra": "^10.1.0",
         "rimraf": "^3.0.2",
         "rollup": "3.1.0",
+        "tsx": "^3.12.1",
         "typescript": "4.8.4"
       },
       "engines": {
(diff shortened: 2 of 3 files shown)


.github/workflows/ci.yml
```diff
@@ -30,5 +30,4 @@ jobs:
         run: bash <(curl -s https://codecov.io/bash) -t ${{secrets.CODECOV_TOKEN}} -B ${{ github.ref }} -f coverage/coverage-final.json
       # test the minimum supported peer dependency version
       - run: npm install typescript@4.1 rollup@3.0
-      # aka `npm test` without the `pretest/build`
-      - run: node .build/tests/index.js
+      - run: npm test
```
Swatinem, line 33: I’m surprised this does not fail, skipping the build step was necessary because building with an outdated TS version used to fail.

Swatinem (commented):
TBH, I would rather not do this.

Splitting the build from the test/run steps was intentional. We have to typecheck things anyway using `tsc` and might as well use its output. Otherwise you have to typecheck twice which is useless overhead.

I’ve had rather suboptimal experience with the various "run tsx directly" tools over the years.

.github/workflows/ci.yml
privatenumber: This PR removes the build step, so this is running the TS test files as if they're plain JS by stripping out the types using esbuild.

privatenumber:
> Otherwise you have to typecheck twice which is useless overhead.

If this PR were to be merged, I was planning to open a PR to update the build to use [pkgroll](https://github.com/privatenumber/pkgroll) (no-config Rollup):
https://github.com/privatenumber/rollup-plugin-dts/compare/tests-tsx...privatenumber:rollup-plugin-dts:pkgroll

The advantage is both building and test-running will be instantaneous. During local development, type-checking will happen in the background via IDE. When pushed, type-checking gets verified via `npm run type-check` on CI.

Feel free to close if this is still something you prefer to avoid. Thanks for the feedback regardless!

Swatinem:
I would rather prefer to keep the dependencies to a minimum.

I remember TS used to have something to emit files as the IDE was running, though I never got that to work back in the days. That would at least make testing instantaneous if you decouple the test step from the build step.

privatenumber:
No problem! Thanks for looking into it

PR #249, opened 2023-02-12
properly support watch mode when transforming .ts files

### LINKED ISSUE #248: watch mode does not watch for changes to files when using indirectly inferred types
### Checklist

* [ ] I am running this plugin on `.d.ts` files generated by TypeScript.

### Code Snipped

a minimum reproduction repo can be found here: https://github.com/rayzr522/rollup-plugin-dts-watch-bug

```ts
// index.ts
export type { MyRouter as test } from "./types";

// types.ts
import { myRouter } from "./router";

const x = myRouter;
export type MyRouter = typeof x;

// router.ts
export const myRouter = { someProcedure: () => 2 + 2 };
```

### Issue

when exporting a type that is inferred indirectly from an import, watch mode does not trigger a rebuild when changing the indirectly referenced file. this can be seen in the following demo video:

https://user-images.githubusercontent.com/15263431/218290565-10a892cd-ff73-40c8-8b64-f46615ec82fe.mp4

the code used in this video is also included in my minimal repro repo: https://github.com/rayzr522/rollup-plugin-dts-watch-bug

I understand using this plugin directly on .ts files isn't necessarily recommended, but it seems like it should be relatively trivial to tell Rollup to watch every file that ends up being handled by this plugin. i did a little digging to see if I could open a quick PR, but didn't get very far before realizing it was probably most efficient to just open an issue report and give all the context I've gained so far.

as far as I understand, since you're using the typescript compilation API programmatically, it should be somewhat trivial to get a list of files that ended up being imported, and then perhaps using [`addWatchFile`](https://rollupjs.org/plugin-development/#this-addwatchfile) to tell rollup to watch them?

this bug makes it a huge pain in the ass when trying to bundle a client package for a [tRPC](https://trpc.io/) router, as the idea is that you define your tRPC router in-code, take the `typeof` that router, and export that in your client package so that consumers can have a type-safe client without you having to bundle all your server-side-only router code into your client bundle

in the interim, I am just going to add an unused direct export of my sub-routers so that watch mode picks them up, but this feels like 100% unintentional behavior

happy to help in any way I can! :3

rayzr522:
closes #248

while the diff looks noisy, these changes preserve the codepaths 1:1 while making the actual logic involved much more digestible

the main big change is in resolving modules, where we now resolve both the `code` as well as the typescript-compiler-specific `source` & `program` values. this makes the logic inside of `transform()` way simpler, where we can just check that `resolveModule` actually returned something, and if so we know we can pass the code straight along to the inner transform plugin

I also did my best to remove a lot of mutability in variables as it was making it hard to follow the logic, as well as make the types a little stricter/more explicit

if this PR is too heavy-handed, I am happy to restore some of the original structure (e.x. re-inlining getModule so we don't need to pass around a context object) while preserving the main core changes that enabled better watch mode support. however, as you mentioned in #248 that this repo is mostly in maintenance mode, I am hoping that these changes will be welcome as I feel they greatly improve the readability & robustness of the code :3

all tests pass :)


rayzr522:
the changes to use `satisfies` were mainly so that we know the exact hooks/functions in each plugin, eliminating the need to handle the nullability of the transform plugin's hooks since typescript now knew that the functions we required were 100% present

I'm all for exposing the main plugin with an explicit typedef if you don't want to have the public side of things to have those explicit typings, but at least for the inner transform plugin, it was hugely helpful to be able to have more explicit typings that mirrored the actual code

tysm for getting this released!

PR #250, opened 2023-02-16
Allow to use same resolve method as tsconfig (e.g. NodeNext)

remcovaes:
While trying to inline external dependencies, I saw that a TypeScript function was used that did not respect the moduleResolution of the config. It is changed to a similar functions that does respect the moduleResolution. There is also a test added that verifies that it works for NodeNext/Node16 for moduleResolution.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -193,7 +193,7 @@ export default function rollupPluginDts(options: Options = {}) {
       }
 
       // resolve this via typescript
-      const { resolvedModule } = ts.nodeModuleNameResolver(source, importer, resolvedCompilerOptions, ts.sys);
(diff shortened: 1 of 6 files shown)


Swatinem:
Looks like you have to add `tsVersion: "x.y",` to the `meta` file, as this seems to be failing with the minimum supported TS 4.1.

Swatinem (approved):

riovir:
This PR is much appreciated! It enables us to turn an otherwise breaking change (that requires our users to change their TypeScript and project setup along the lines of module: NodeNext and type: module) of our design system into a simple patch version. Can't wait for a release. :)

PR #253, opened 2023-02-27
Add support for using `.d.cts` and `.d.mts` as inputs

### LINKED ISSUE #252: Allow using `.d.cts` and `.d.mts` files as input
### Checklist

### Code Snipped

[Sample reproduction](https://github.com/Comandeer/bugs-reports/tree/rollup-plugin-dts-cts-mts).

Newer TypeScript versions by default generate a `.d.mts` file for an `.mts` file and a `.d.cts` file for a `.cts` file. #228 added support for `.mjs` and `.cts `files and #196 – support for generating `.d.cts` and `.d.mts` files. However, this plugin does not support using `.d.cts` and `.d.mts` files as inputs, and it seems like a missing piece of support for new TS file extensions.

Currently, all `.d.cts` and `.d.mts` files passed as input are ignored by the plugin, resulting in Rollup throwing an error about unsupported syntax. They need to be renamed to `*.d.ts` in order to be parsed and bundled.

Sample Rollup config:

```ts
import dts from 'rollup-plugin-dts';

export default {
	input: {
		cts: './dist/cjsModule.d.cts',
		mts: './dist/esmModule.d.mts'
	},
	output: {
		dir: './types'
	},
	plugins: [
		dts()
	]
};
```

### Error Message

```
[!] RollupError: Unexpected token (Note that you need plugins to import files that are not JavaScript)
dist/esmModule.d.mts (1:8)
1: declare function mjsModule(): string;
           ^
2: export default mjsModule;
    at error (/home/comandeer/repos/other/bugs-reports/node_modules/rollup/dist/shared/rollup.js:261:30)
    at Module.error (/home/comandeer/repos/other/bugs-reports/node_modules/rollup/dist/shared/rollup.js:13642:16)
    at Module.tryParse (/home/comandeer/repos/other/bugs-reports/node_modules/rollup/dist/shared/rollup.js:14319:25)
    at Module.setSource (/home/comandeer/repos/other/bugs-reports/node_modules/rollup/dist/shared/rollup.js:13929:39)
    at ModuleLoader.addModuleSource (/home/comandeer/repos/other/bugs-reports/node_modules/rollup/dist/shared/rollup.js:23894:20)
```

Comandeer:
I've introduced the `DTS_EXTENSIONS` regex, similar to the existing `TS_EXTENSIONS` one, and used it in places that were simply checking if the file name ends with the `.d.ts`. Thanks to that `.d.cts` and `.d.mts` files can be used as inputs for the plugin.

Closes #252.

### DIFF
--- a/src/program.ts
+++ b/src/program.ts
@@ -1,6 +1,7 @@
 import * as path from "path";
 import ts from "typescript";
 
+export const DTS_EXTENSIONS = /\.d\.(c|m)?tsx?$/;
 export const dts = ".d.ts";
 
 export const formatHost: ts.FormatDiagnosticsHost = {
@@ -91,7 +92,7 @@ export function getCompilerOptions(
   }
   const { fileNames, options, errors } = configByPath.get(cacheKey)!;
 
-  dtsFiles = fileNames.filter((name) => name.endsWith(dts));
+  dtsFiles = fileNames.filter((name) => DTS_EXTENSIONS.test(name));
   if (errors.length) {
     console.error(ts.formatDiagnostics(errors, formatHost));
     return { dtsFiles, dirName, compilerOptions };
@@ -123,7 +124,7 @@ export function createPrograms(input: Array<string>, overrideOptions: ts.Compile
   let compilerOptions: ts.CompilerOptions = {};
 
   for (let main of input) {
-    if (main.endsWith(dts)) {
+    if (DTS_EXTENSIONS.test(main)) {
       continue;
     }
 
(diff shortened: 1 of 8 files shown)


src/program.ts
```diff
@@ -1,6 +1,7 @@
 import * as path from "path";
 import ts from "typescript";
 
+export const DTS_EXTENSIONS = /\.d\.(c|m)?tsx?$/;
 export const dts = ".d.ts";
```
Swatinem, line 5: is this still in use now?

Swatinem (approved):

src/program.ts
Comandeer: Yes, it's used during generation of `.d.ts` file from TS → https://github.com/Comandeer/rollup-plugin-dts/blob/203d22334022eac109a63bbe9c00386d7e8532df/src/index.ts#L143

Issue #162, opened 2021-08-01
UnsupportedSyntaxError: namespace child (hoisting) not supported


skaneprime:
### Checklist

* [x] I can reproduce this issue when running this plugin on its own.
  *Other plugins, such as `node-resolve` are known to cause issues.*
* [x] I am running this plugin on `.d.ts` files generated by TypeScript.
  *The plugin can consume `.ts` and even `.js` files (with `allowJs: true`), but this is known to cause issues.*
* [x] This issue is not related to rolling up `@types`.
  *The plugin ignores these by default, unless `respectExternal` is set. `@types` can contain hand-crafted code which is known to cause issues.*

### Code Snipped

```ts
// Client.ts
import * as Mods from './modules'

export class Client {
    Services = new Mods.Services()
}
export namespace Client {
    export import Services = Mods.Services
}
```

```ts
// ./modules/index.ts
export * from './Services'
```

```ts
// ./modules/Services.ts
export namespace Services {
    export type Type = "ServiceType"
}
export class Services {}
```

### Error Message

```ts
// The Error Message thrown by rollup / the plugin.
Failed to build: UnsupportedSyntaxError: namespace child (hoisting) not supported yet
   6 | declare namespace Client {
>  7 |     export import Services = Mods.Services;
     |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   8 | }
   9 |
  10 | export { Client };
```

Swatinem:
I experimented with fixing this, but rollup flattens all the imports into the global namespace, and typescript does not allow re-declarations, which is super annoying, as this generated code fails in typescript:

```
declare namespace Services {
  type Type = "ServiceType";
}
declare class Services {}
declare class Client {}
declare namespace Client {
  export import Services = Services;
}
export { Client };
```

skaneprime:
@Swatinem If it flattens all imports then is there a way to redeclare a namaspace? 
```ts
// Child.ts
export namespace Child {}
// Main.ts
import { Child as ChildTemp } from './Child.ts'
export namespace Main {
     export import Child = ChildTemp
}
```

Swatinem:
Maybe there is, but I doubt its worth the effort.

Danielku15:
I was about to file a similar issue and during my tests I even found a simpler scenario without namespaces not being supported. Am I right with my assumption that these problems relate I guess solving the following issue, should also fix the exports within namespaces:

```ts
[...]

PR #261, opened 2023-04-24
add support for import aliases

### LINKED ISSUE #246: Nested namespace identifiers errors with `namespace must have a "ModuleBlock" body`
### Checklist

### Code Snipped

```ts
declare namespace Foo.Bar {}
```

### Error Message

```
Error: namespace must have a "ModuleBlock" body.
```

### Details

I get this error when using the shorthand for nesting namespaces (separating identifiers with dots).

I believe this is a more general case of https://github.com/Swatinem/rollup-plugin-dts/issues/160.

Looking at [the AST for this](https://ts-ast-viewer.com/#code/CYUwxgNghgTiAEA7KBbEBnADlMCBiA9gQHQBCs8A3gL4BQQA), the `body` of the first `ModuleDeclaration` is indeed another `ModuleDeclaration` and not a `ModuleBlock` as [expected by the plugin](https://github.com/Swatinem/rollup-plugin-dts/blob/v5.1.1/src/transform/DeclarationScope.ts#L336).

Patching the [default value of `relaxedModuleBlock `](https://github.com/Swatinem/rollup-plugin-dts/blob/v5.1.1/src/transform/DeclarationScope.ts#L328) to `true` seems to solve the problem, but what would the drawback of that be?

rayzr522:
i've been blocked with this plugin due to the lack of support for import aliases (i.e. `export import something = Some.Namespace.Member`) in a project at work. I'm not sure if this is the correct solution, as I will be honest I am not actually sure why my changes in Transformer cause it to work. i think I am at the limits of my understanding of this codebase. removing either the `.createDeclaration` or the `.pushStatement` causes the import aliases to be dropped from the output, but I don't really understand why.

nevertheless, this does fix support and I have added test cases to prove it. i also fixed nested namespace support while I was at it as some of my own personal tests involved a nested namespace.

also closes #246, although that was not my original purpose :P

I think it might be a step in the right direction for some of the ppl over in #162 but only a couple of their cases seem to be related to import aliases. the majority seem to genuinely be related to hoisting children of namespaces

please advise if you have any better ideas for how to support this! all existing tests pass as well as the new cases I added in order to confirm support.

### DIFF
--- a/src/transform/DeclarationScope.ts
+++ b/src/transform/DeclarationScope.ts
@@ -346,7 +346,8 @@ export class DeclarationScope {
         ts.isClassDeclaration(stmt) ||
         ts.isInterfaceDeclaration(stmt) ||
         ts.isTypeAliasDeclaration(stmt) ||
-        ts.isModuleDeclaration(stmt)
+        ts.isModuleDeclaration(stmt) ||
+        ts.isImportEqualsDeclaration(stmt)
       ) {
         if (stmt.name && ts.isIdentifier(stmt.name)) {
           this.pushTypeVariable(stmt.name);
@@ -407,6 +408,10 @@ export class DeclarationScope {
         // noop
         continue;
       }
+      if (ts.isImportEqualsDeclaration(stmt)) {
+        // noop
+        continue;
+      }
       if (ts.isExportDeclaration(stmt)) {
         if (stmt.exportClause) {
           if (ts.isNamespaceExport(stmt.exportClause)) {
@@ -417,9 +422,9 @@ export class DeclarationScope {
             this.pushIdentifierReference(id);
           }
         }
-      } else {
-        throw new UnsupportedSyntaxError(stmt, `namespace child (walking) not supported yet`);
+        continue;
       }
+      throw new Error(`unsupported statement type ${ts.SyntaxKind[stmt.kind]}`);
     }
 
     this.popScope();
--- a/src/transform/DeclarationScope.ts
+++ b/src/transform/DeclarationScope.ts
@@ -325,11 +325,11 @@ export class DeclarationScope {
     }
   }
 
-  convertNamespace(node: ts.ModuleDeclaration, relaxedModuleBlock = false) {
+  convertNamespace(node: ts.ModuleDeclaration) {
     this.pushScope();
 
-    if (relaxedModuleBlock && node.body && ts.isModuleDeclaration(node.body)) {
-      this.convertNamespace(node.body, true);
+    if (node.body && ts.isModuleDeclaration(node.body)) {
+      this.convertNamespace(node.body);
       return;
     }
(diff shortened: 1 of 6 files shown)


src/transform/Transformer.ts
```diff
@@ -248,28 +248,50 @@ class Transformer {
     }
   }
 
+  convertNamespaceAlias() {}
+
```
florrdv, line 252: Did you mean to leave this in?

rayzr522: oops, nope. thanks for catching that!

src/transform/DeclarationScope.ts
```diff
@@ -400,13 +401,17 @@ export class DeclarationScope {
...
+        this.convertNamespace(stmt);
         continue;
       }
       if (ts.isEnumDeclaration(stmt)) {
         // noop
         continue;
       }
+      if (ts.isImportEqualsDeclaration(stmt)) {
+        // noop
+        continue;
```
Swatinem, line 413: I believe you need to convert the right hand side here as well.

src/transform/Transformer.ts
```diff
@@ -251,25 +251,45 @@ class Transformer {
...
+              source: convertExpression(node.moduleReference.expression) as any,
+            },
+            node,
+          ),
+        );
+        return;
+      } else if (ts.isQualifiedName(node.moduleReference)) {
+        // TODO: I *think* this just ensures that the statement doesnt get tree-shaken out of existence, even though there's no real variable declaration here that makes it into the final build
+        this.createDeclaration(node, node.name);
+        this.pushStatement(
```
Swatinem, line 274: This indeed feels a bit out of place here. Maybe merging this with the above `if` will also work, not sure though.

Swatinem (commented):
You can set a special env var to dump the fake-AST that this plugin generates:

https://github.com/Swatinem/rollup-plugin-dts/blob/f74e5a972cc613960aead51c4ff3415a7b3dce83/src/transform/index.ts#L86-L90

That AST should resemble "normal" JS code that is being processed by rollup. This always helps debugging these things.

In general this looks good. I would however add a couple more tests to this. In particular, can you add a `export import Foo = Bar` where `Bar` is in the global scope, and would otherwise be dead-code-eliminated if not for the reference through `Foo`. I believe that should cover my comment about converting the right hand side of the import equals declaration.

ClementNerma:
Hi, is there anything new here? I'm currently blocked because of the lack of support for the `export import` syntax, and I see there wasn't any activity synce April 25th. Is this PR still active?

rayzr522:
i have long since moved on from the project which was using this, and i felt shakey about this PR to begin with since i truly did not understand how a lot of the black magic in this plugin worked lol. but i will leave my fork around so my changes can be reviewed in case anyone else wants to revive this effort!

PR #274, opened 2023-07-23
refactor: change plugin export from default to named

### LINKED ISSUE #247: TypeError: dts is not a function
### Checklist

### Notes

- I'm following https://plainenglish.io/blog/react-component-library-with-ts-and-rollup-77aca0c727bf#rollupconfigjs 
- https://stackoverflow.com/a/74304876 fixes the issue, but I thought I would log the issue since the README suggests the approach in the blog post I was following.

### Dependency versions

```
"devDependencies": {
    "@rollup/plugin-commonjs": "^24.0.1",
    "@rollup/plugin-node-resolve": "^15.0.1",
    "@rollup/plugin-terser": "^0.4.0",
    "@rollup/plugin-typescript": "^11.0.0",
    "@types/react": "^18.0.27",
    "classnames": "^2.3.2",
    "react": "^18.2.0",
    "rollup": "^3.14.0",
    "rollup-plugin-dts": "^5.1.1",
    "rollup-plugin-peer-deps-external": "^2.2.4",
    "rollup-plugin-postcss": "^4.0.2",
    "tslib": "^2.5.0",
    "typescript": "^4.9.5"
  }
```

### Code Snipped

```ts
import resolve from "@rollup/plugin-node-resolve";
import commonjs from "@rollup/plugin-commonjs";
import typescript from "@rollup/plugin-typescript";
import terser from "@rollup/plugin-terser";
import external from "rollup-plugin-peer-deps-external";
import postcss from "rollup-plugin-postcss";
import dts from "rollup-plugin-dts";

const packageJson = require("./package.json");

export default [
  {
    input: "src/index.ts",
    output: [
      {
        file: packageJson.main,
        format: "cjs",
        sourcemap: true,
        name: 'kingdom-ui',
      },
      {
        file: packageJson.module,
        format: "esm",
        sourcemap: true,
      },
    ],
    plugins: [
      external(),
      resolve(),
      commonjs(),
      typescript({ tsconfig: "./tsconfig.json" }),
      postcss(),
      terser(),
    ],
  },
  {
    input: "dist/esm/types/index.d.ts",
    output: [{ file: "dist/index.d.ts", format: "esm" }],
    external: [/\.css$/],
    plugins: [dts()],
  },
];

```

### Error Message

```
[!] TypeError: dts is not a function
```

spiltbeans:
## Describing the problem
If you import the plugin as directed in the docs you may encounter a console error like:
```js
import dts from "rollup-plugin-dts";

const config = [
  // …
  {
    input: "./my-input/index.d.ts",
    output: [{ file: "dist/my-library.d.ts", format: "es" }],
    plugins: [dts()],
  },
];

export default config;
```

```
[!] TypeError: dts is not a function
```

And you will be returned this if you log `dts` to the console:
```
{ default: [Function: plugin] }
```
This PR closes #247.

## Describing the solution
This PR modifies the export of the `plugin` function as a named export called `dts` rather than `export default`.

Using named exports is less likely to encounter these inconsistent importing patterns in the future.

This [StackOverflow thread](https://stackoverflow.com/questions/46913851/why-and-when-to-use-default-export-over-named-exports-in-es6-modules) provides multiple reasons why named exports are a favourable pattern for ES6. [This response](https://stackoverflow.com/a/68665805) compiles relevant articles expanding on the issue.

## Describing changes to the package
This PR would change the way this package is imported into projects. It will be a breaking change as importing will now be:

```js
// new import
import { dts } from "rollup-plugin-dts";

// example usage
const config = [
  // …
  {
    input: "./my-input/index.d.ts",
    output: [{ file: "dist/my-library.d.ts", format: "es" }],
    plugins: [dts()],
  },
];

export default config;
```

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -209,4 +209,4 @@ const plugin: PluginImpl<Options> = (options = {}) => {
   } satisfies Plugin;
 };
 
-export { plugin as dts } ;
+export { plugin as dts, plugin as default } ;
--- a/src/index.ts
+++ b/src/index.ts
@@ -209,4 +209,4 @@ const plugin: PluginImpl<Options> = (options = {}) => {
   } satisfies Plugin;
 };
 
-export default plugin;
+export { plugin as dts } ;
--- a/tests/testcases.ts
+++ b/tests/testcases.ts
@@ -10,7 +10,8 @@ import {
   VERSION as rollupVersionMajorMinorPatch,
 } from "rollup";
 import ts from "typescript";
-import dts, { type Options } from "../src/index.js";
+import { dts } from "../src/index.js";
+import { type Options } from "../src/index.js";
 import { exists, forEachFixture, Harness } from "./utils.js";
 
 export default (t: Harness) => {
--- a/tests/testcases.ts
+++ b/tests/testcases.ts
@@ -10,8 +10,7 @@ import {
   VERSION as rollupVersionMajorMinorPatch,
 } from "rollup";
 import ts from "typescript";
-import { dts } from "../src/index.js";
-import { type Options } from "../src/index.js";
+import dts, { type Options } from "../src/index.js";
 import { exists, forEachFixture, Harness } from "./utils.js";
 
 export default (t: Harness) => {
--- a/README.md
+++ b/README.md
@@ -28,6 +28,8 @@ const config = [
 export default config;
 ```
 
+**NOTE** A [default import](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/import#default_import) of the plugin using `import dts from "rollup-plugin-dts";` is still supported for existing implementations of this package. However, a [named import](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/import#named_import) is suggested to avoid the error `[!] TypeError: dts is not a function`[^1][^2] in certain rollup config file implementations.
+
 And then instruct typescript where to find your definitions inside your `package.json`:
 
 ```json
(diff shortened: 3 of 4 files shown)


spiltbeans:
I've created an alternate PR in #275 to solve the same issue. They are different solutions and only one should be accepted.

src/index.ts
```diff
@@ -209,4 +209,4 @@ const plugin: PluginImpl<Options> = (options = {}) => {
   } satisfies Plugin;
 };
 
-export default plugin;
+export { plugin as dts } ;
```
Swatinem, line 212: you can keep both the named export and the default export I believe

Swatinem (approved):
Thanks for taking care of this. I would be curious about what part of the commonjs interop exactly is breaking this usecase here

src/index.ts
spiltbeans: If the plugin is both a named export and a default export, which should be documented in the README.md?

Maybe a hybrid approach with a note like #275 to say "if you're getting this error you can do a restructured import"?

Swatinem: I believe you can document the newer named export in the README. But having the default export is rather a backwards compatibility thing, so existing users are not forced to upgrade.

Swatinem: Also, having a comment such as that is still a good idea :-)

spiltbeans: - supporting default export in eeb269a4500d4d3d4a1c416709dbd3809cd93741
- add a note in README in cb57b11c883601477db0b997991b0190968733fa

tests/testcases.ts
```diff
@@ -10,7 +10,8 @@ import {
   VERSION as rollupVersionMajorMinorPatch,
 } from "rollup";
 import ts from "typescript";
-import dts, { type Options } from "../src/index.js";
+import { dts } from "../src/index.js";
+import { type Options } from "../src/index.js";
```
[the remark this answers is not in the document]
spiltbeans: - inline import in 461ec5fddcb3915fc479ef86492adc86fd92465c

Swatinem (approved):

Swatinem:
Thanks again for looking into this! I will get a new release out during this week.

Issue #276, opened 2023-07-26
Invalid exports when using isolatedModules: true


mithodin:
### Checklist

* [x] I can reproduce this issue when running this plugin on its own.
  *Other plugins, such as `node-resolve` are known to cause issues.*
* [x] I am running this plugin on `.d.ts` files generated by TypeScript.
  *The plugin can consume `.ts` and even `.js` files (with `allowJs: true`), but this is known to cause issues.*
* [x] This issue is not related to rolling up `@types`.
  *The plugin ignores these by default, unless `respectExternal` is set. `@types` can contain hand-crafted code which is known to cause issues.*

### Issue Description
When creating files using the tsconfig setting `isolatedModules: true`, type exports must explicitly use `export type { ... }` or `export { type ... }` (TS1205 is the relevant error code). `tsc` does this correctly, e.g. the following code snippet:
```typescript
type Test = {
    challenge: string;
}
export { type Test };
```
is reproduced as-is in the corresponding .d.ts file.
However, rollup-plugin-dts elides the `type` part of the export, leading to the mentioned error.
A minimal reproduction can be found here: https://github.com/mithodin/dts-bundle-isolation-repro
Compile using `npm install && npm run build` and check the file `/dist/bundle.d.ts`. For comparison with tsc, run `npx tsc` and check `/out/gameLibrary.d.ts`.

It should be mentioned that this causes issues for anyone using a library bundled using rollup-plugin-dts, who wants to enable isolatedModules.

mithodin:
The problem persists with version 5.3.1., I have updated my example repo: https://github.com/mithodin/dts-bundle-isolation-repro

Swatinem:
I was trying to bulk-close all of the old issues in the repo, as I have moved the project into maintenance mode: https://github.com/Swatinem/rollup-plugin-dts/issues/277 Too bad github does not allow bulk-close-with-comment.

That being said, I will most likely not work on resolving this issue, but would very much appreciate a PR.

Solving this would likely involve changing the "post processing" code to split type-only exports from normal exports.

PR #278, opened 2023-08-04
Transform type exports to "export type" in order to comply with isolatedModules

mithodin:
Fixes https://github.com/Swatinem/rollup-plugin-dts/issues/276

### DIFF
--- a/src/transform/index.ts
+++ b/src/transform/index.ts
@@ -4,6 +4,7 @@ import ts from "typescript";
 import { NamespaceFixer } from "./NamespaceFixer.js";
 import { preProcess } from "./preprocess.js";
 import { convert } from "./Transformer.js";
+import {ExportsFixer} from "./ExportsFixer.js";
 
 function parse(fileName: string, code: string): ts.SourceFile {
   return ts.createSourceFile(fileName, code, ts.ScriptTarget.Latest, true);
@@ -91,8 +92,8 @@ export const transform = () => {
       return { code, ast: converted.ast as any, map: preprocessed.code.generateMap() as any };
     },
 
-    renderChunk(code, chunk, options) {
-      const source = parse(chunk.fileName, code);
+    renderChunk(inputCode, chunk, options) {
+      const source = parse(chunk.fileName, inputCode);
       const fixer = new NamespaceFixer(source);
 
       const typeReferences = new Set<string>();
@@ -120,7 +121,7 @@ export const transform = () => {
         }
       }
 
-      code = writeBlock(Array.from(fileReferences, (ref) => `/// <reference path="${ref}" />`));
+      let code = writeBlock(Array.from(fileReferences, (ref) => `/// <reference path="${ref}" />`));
       code += writeBlock(Array.from(typeReferences, (ref) => `/// <reference types="${ref}" />`));
       code += fixer.fix();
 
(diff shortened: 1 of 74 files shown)


tests/testcases/computed-property/expected.d.ts
```diff
@@ -10,4 +10,4 @@ type Klass = {
   [0]: C;
   [Dprop]: D;
 };
-export { Klass };
+export type { Klass };
```
Swatinem, line 13: I’m a bit surprised why `A, B, C, D` are not exported in this case 🤔 
But this existed before, so not worth taking a look in this PR

src/transform/index.ts
```diff
@@ -120,15 +121,17 @@ export const transform = () => {
...
+      let code = writeBlock(Array.from(fileReferences, (ref) => `/// <reference path="${ref}" />`));
       code += writeBlock(Array.from(typeReferences, (ref) => `/// <reference types="${ref}" />`));
       code += fixer.fix();
 
       if (!code) {
         code += "\nexport { }";
       }
 
-      return { code, map: { mappings: "" } };
+      const exportsFixer = new ExportsFixer(parse(chunk.fileName, code));
```
Swatinem, line 132: Now the `renderChunk` hook will parse the generated code twice. I believe its fine for now, but might be worth to merge these two post-processors into a single pass.

Swatinem (approved):
This looks like a pragmatic solution that works, great job!

src/transform/index.ts
mithodin: Yes, I thought about doing it in one pass, but ultimately decided that it is cleaner to separate the postprocessing.

mithodin:
@Swatinem it seems that export { type T } is only supported since Typescript 4.5. Would you be willing to raise the minimum compatible ts version to 4.5, or would you prefer we split type exports from value exports? This also fixes the issue with
```
TypeError: Cannot read properties of null (reading 'deoptimizePath')
```
that we see in the ci logs.

tests/testcases/computed-property/expected.d.ts
mithodin: Should they be? They are not exported in the index.d.ts.

mithodin:
Also, support for node 14 can probably be dropped since it is end-of-life since May of this year.

mithodin:
This would probably mean a 4.0.0 release, right?

Swatinem (commented):
I do see both sides of bumping version requirements.

From a consumer point of view, bumping versions in most cases is as easy as a oneliner. If it isn’t, and someone is stuck on older versions, it does not make a difference being stuck on an older version of this plugin either.

From a maintainers point of view, lowering the maintenance burden by not having to support older versions is definitely a plus. Given than backwards compatibility to older Rollup versions was already broken by accident recently, might as well bump that up as well in accordance with that. For the Typescript version itself, not sure what a reasonable version requirement there would be. `4.5` sounds reasonable, as that was released end of 2021. Might as well bump higher, as I really don’t see a point supporting TS older than a year.

.github/workflows/ci.yml
```diff
@@ -6,7 +6,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        node: [14, 18]
+        node: [16, 18, 20]
```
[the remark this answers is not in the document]
mithodin: done, changed minimum version to node 18

mithodin:
@Swatinem thanks for the merge :-) Do you have any plans yet for when to do a release?

mithodin:
Hi @Swatinem 
It would be really cool if you could release this so we can actually get our issue fixed :-) Could you please do it?

Swatinem:
Done, sorry for the delay.

Issue #282, opened 2023-08-24
Modify engines.node>=16


curlykay9527:
When using the "engine-strict = true" configuration in .npmrc, it is easy to encounter installation failures for version 6.0.0 if the latest LTS version of Node is not used.
Due to the project entering a maintenance state, unless there is a specific requirement to use Node version 18.7.1, it is recommended to set the minimum required Node version to >=16.

Swatinem:
I’m happy to take a PR that lowers the requirements to >=16.

curlykay9527:
`renovate[bot]` will regularly modify engines.node to the latest version 3cead14ada32de3af636b604efdf4dcf486a2072, which makes the pr #283  invalid.
hope to modify the dependency update strategy of `renovate [bot]` to ignore updates for engines.node.

Swatinem:
Either one of those should be fine really.
I will close this issue for now, as https://github.com/Swatinem/rollup-plugin-dts/blob/master/CHANGELOG.md#601 was released with a version set to 16. Happy to take further PRs to avoid renovate messing with this, it has been annoying me as well.

PR #283, opened 2023-08-25
chore(deps): lower the node requirements to >=16

curlykay9527:
lower the `node` requirements to `>=16`,reduce install fail caused by the `engine-strict = true` configuration item in `.npmrc`. #282

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -6,7 +6,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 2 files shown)


Swatinem (approved):
can you also change CI to test with that version please?

curlykay9527:
> can you also change CI to test with that version please?

no problem it should be

Swatinem (approved):

elchininet:
When will be released a new version with this change?

PR #285, opened 2023-08-29
fix: treat an exported namespace as a value export

### LINKED ISSUE #284: Namespaced type is exported as "export type" instead of just "export"
### Checklist

### Code Snipped

```ts
export interface MyInterface {
  // stuff
}

export namespace MyInterface {
  // stuff
}
```

### Produced Result

In 6.0.0 the output is this

```tsx
interface MyInterface {
  // stuff
}
declare namespace MyInterface {
  // stuff
}

export type { MyInterface };
```

When used in other projects, this error is thrown:

`'MyInterface' cannot be used as a value because it was exported using 'export type'.`

### Expected Result

In 5.3.0 the result is this

```tsx
interface MyInterface {
  // stuff
}
declare namespace MyInterface {
  // stuff
}

export { MyInterface };
```

mithodin:
fixes #284


Swatinem (approved):
thanks for the quick turnaround :-)

PR #324, opened 2024-11-20
Feat/project references

NWYLZW:


NWYLZW:
All tests for Windows have failed. The apparent reason is also puzzling.

Swatinem:
the actual modifications look good. though I’m not a big fan of pulling in `vitest` for this. is it possible to integrate this into the existing testsuite?

NWYLZW:
> the actual modifications look good. though I’m not a big fan of pulling in `vitest` for this. is it possible to integrate this into the existing testsuite?

I also tried to directly use the current testing tool. However, at first, I didn't know how to use the current testing tool very well, so I first used `vitest` for testing work.

PR #333, opened 2025-02-06
fix: imports and exports follows type-only

### LINKED ISSUE #311: type only exports of classes are not respected
### Checklist

### Code Snippet

```ts
declare class Foo {
  bar: string;
}

export type { Foo };
```

#### Actual Output

```ts
declare class Foo {
  bar: string;
}
export { Foo };
```

#### Expected Output

```ts
declare class Foo {
  bar: string;
}
export type { Foo };
```

### Additional Context

It works if I alias my class as a type instead:

```ts
declare class _Foo {
  bar: string;
}

type Foo = _Foo;

export type { Foo };
```

#### Output

```ts
declare class _Foo {
  bar: string;
}
type Foo = _Foo;
export type { Foo };
```

kricsleo:
resolves #311

I've been working on an approach that might seem a bit unconventional and wondering if you could review and provide feedback on these implementation details:

1. Preprocessing source code by renaming and adding type aliases to preserve import/export variable type information in variable names

2. Restoring import/export variable type information in the `renderChunk` phase based on the variable names containing type information

Would love to hear your thoughts on potential improvements or simpler alternatives. 
Are there any concerns or better approaches you can suggest?


Swatinem (approved):
looks reasonable to me, thanks a lot :-)

kricsleo:
If a new version is to be released, I would suggest an rc version. 🤔
Because I'm not sure if this will cause any unintended disruptive changes (due to many edge cases).

PR #336, opened 2025-02-19
feat: support importing json modules

### LINKED ISSUE #184: Rollup fails if I import a json file
### Checklist

### Code Snipped

```ts
import v8 from './reference/v8.json';
```

### Error Message

```
[!] Error: Could not resolve './reference/v8.json' from types/src/style-spec/style-spec.d.ts
```

The long story:
I've recently migrated a javascript library written in flow to typescript.
This library had a definition file before the typescript migration and everything worked fine:
https://github.com/maplibre/maplibre-gl-js/blob/cb3a4200bfbc113969e9dcbedae2aa7655f9aa98/src/index.d.ts
After the typescript migration I saw that the generated `d.ts` file are not producing the same output (not even remotely close).
I saw that this rollup plugin might help me with my task of creating a single d.ts file generated from the typescript code, but I'm not sure how easy this will be as the typescript code is a bit complicated.
The code can be found here:
https://github.com/maplibre/maplibre-gl-js/pull/623
I'm working on a branch but the error I get is for a local attempt I'm doing in order to see what this plugin produces

### LINKED ISSUE #332: Incorrect d.ts When using the json plugin and reexporting a json file
I open an stackblitz playground to review the bug:

https://stackblitz.com/edit/stackblitz-starters-spet46kk?file=dist%2Findex.d.ts

When I write `export { version } from './package.json'`, the resulting d.ts file looks like this:

```ts
var version = "0.0.1";

export { version };
```

It have two errors:

Image

```

Top-level declarations in .d.ts files must start with either a 'declare' or 'export' modifier.(1046)
```
Image

```
Initializers are not allowed in ambient contexts.(1039)
```

In fact, the code we need to get is this:

Image

```ts
declare var version: '0.0.1';

export { version };
```

kricsleo:
resolves #332
resolves #184

(Can also resolve some downstream issues: https://github.com/unjs/unbuild/issues/430)


kricsleo:
This is so strange that the CI test has been returning inconsistent results for `testcases/import-json`. 
Sometimes the result is [`declare let name: ...`](https://github.com/Swatinem/rollup-plugin-dts/actions/runs/13409833405/job/37457243319#step:5:95) and sometimes it's [`declare const name: ...`](https://github.com/Swatinem/rollup-plugin-dts/actions/runs/13409895894/job/37457457058?pr=336#step:8:55). 
I don't have this issue locally.

kricsleo:
Thanks for the detailed explanation ❤️
I tried to bisect and determine that this behaviour was caused by https://github.com/microsoft/TypeScript/pull/53206.

The impacted official version is `v5.1.3`.

- `< v5.1.3`(The closest version: `5.0.4`)： `declare const name: ....`
- `>= v5.1.3`： `declare let name: ....`

I'm not sure if it's meant to change the behavior of JSON modules, but since it doesn't impact actual usage, I'll let it be. 👌🏻

Swatinem (approved):

PR #337, opened 2025-02-25
fix: correctly restore type-only import/export names

### LINKED ISSUE #311: type only exports of classes are not respected

kricsleo:
resolves #311

This is a rework of PR: #333 🫡

- Address more edge cases (see the added testcase)
- Using `AssignmentExpression` to record the type-only hints (Instead of renaming import/export variables previously, since it's unstable and has some minor flaws.)
- Merge `ExportFixer` into `TypeOnlyFixer` since it's now part of `TypeOnlyFixer`'s responsibilities
- Separate type-only and value import/export statements
- Make the type-only fix easier to understand and maintian


Swatinem (approved):

kricsleo:
I’ve got to say, it’s been really tough for me trying to figure out how to implement this feature 🥲🥲

Hope to hear your feedback ❤️

Swatinem (approved):

Swatinem:
To be quite honest, I haven’t really been maintaining this project for years, so I’m not that familiar with the code anymore myself to be of much help here.

But I do trust the testsuite quite a bit, and the results look good to me.

PR #338, opened 2025-03-12
fix: Return Source Map Differences to Rollup

alan-agius4:
This commit enables users to remap declaration source maps by passing back the computed difference.

To utilize the original DTS source map, users may need additional plugins like `rollup-sourcemap`.


Swatinem (approved):

alan-agius4:
@Swatinem, thanks for merging this. Do you have an ETA when to expect the a release?

alan-agius4:
HI @Swatinem, when should we expect a new release? Thanks

PR #348, opened 2025-08-13
fix: Preserve imports for types used in typeof generic 

dayongkr:
##  Problem

When using typeof expressions with generic type parameters from external imports, the plugin incorrectly removed the import statements for those types.

### Example:
``` ts 
 // Input
import { Test } from './models';
import { test } from './test';
export const b: ReturnType<typeof test<Test>> = test({
a: 's' });
```

``` ts
// Output (BEFORE fix)
import { test } from './test';
declare const b: ReturnType<typeof test<Test>>;// Test is undefined!
export { b };
```

``` ts
// Output (AFTER fix) 
import { Test } from './models';// Import preserved
import { test } from './test';
declare const b: ReturnType<typeof test<Test>>;
export { b };
```

## Root Cause

The issue occurred in `DeclarationScope.ts` when processing `TypeQueryNode` (i.e., typeof expressions). The code correctly handled the main expression (`test`) but ignored type arguments (`<Test>`), causing:

1. `typeof test<Test>` was parsed as a `TypeQueryNode` with
  typeArguments
2. Only test was processed as a value reference
3. The type argument Test was not processed, so its import was considered unused
4. `TypeOnlyFixer` removed the "unused" import for Test

## Solution

Added type argument processing to `TypeQueryNode` handling in `DeclarationScope.ts`:

``` ts
  if (ts.isTypeQueryNode(node)) {
    const reference =
  this.convertEntityName(node.exprName);
    this.pushReference(reference);

    if (reference.type === "Identifier") {
      this.markAsValue(reference.name);
    } else if (reference.type === "MemberExpression" &&
  reference.object.type === "Identifier") {
      this.markAsValue(reference.object.name);
    }

    // NEW: Process type arguments in typeof expressions
    this.convertTypeArguments(node);

    return;
  }
```

##  Technical Details

  - File modified: `src/transform/DeclarationScope.ts`
  - Method: `convertTypeNode()`
  - Node type: `TypeQueryNode` (represents typeof expressions)
  - Fix: Added `this.convertTypeArguments(node)` call to process generic type parameters

This ensures that when processing typeof test<Test>, both test (the function) and Test (the type parameter) are properly recognized as references, preventing the import removal.

## Testing

Added test case typeof-generic-type-parameter that verifies:
- :white_check_mark: Import statements for generic type parameters are preserved
- :white_check_mark: Complex type expressions like `ReturnType<typeof func<Type>>` work correctly
- :white_check_mark: Existing functionality remains unaffected


dayongkr:
Due to this issue, function return types can be inferred as any. TS2742 forces us to explicitly set return types in many cases. I hope this gets fixed soon — I'll patch it for now.

tests/testcases/typeof-generic-type-parameter/meta.js
```diff
@@ -0,0 +1,14 @@
...
+export default {
+  tsVersion: "4.8",
+  options: {
+    respectExternal: true,
+  },
+  rollupOptions: {
+    input: {
+      index: "index.ts",
+    },
+    external: ["./models", "./test"],
```
Swatinem, line 12: I would prefer if the test would just work on `.d.ts` files, and bundle everything. Tests that transform `.ts` files are considerably slower as they run the whole typescript compiler to emit intermediate `.d.ts` declarations.

Bundling everything without `external` should also exhibit the original bug, as it would otherwise just dead-code-eliminate the `Test`.

Swatinem (approved):
changes look good to me, though I would still prefer the changes to the test.

tests/testcases/typeof-generic-type-parameter/meta.js
dayongkr: Thank you for the review! I've changed it to a .d.ts file. 

However, this bug only occurs when importing external dependencies, so handling them as external is necessary. In the test environment, If I remove the external setting, it imports fine. For this reason, I chose not to modify the external configuration.

dayongkr:
@Swatinem Could you let me know when the release is planned? That would help me decide whether to patch it for use or simply install the released version. I really appreciate your review.

PR #350, opened 2025-08-19
fix: rewrite names of module declarations with relative paths

### LINKED ISSUE #345: Produces wrong output for declare module statement with relative path
### Checklist

### Code Snipped

See the [minimal reproduction](https://github.com/schoel-bis/rollup-plugin-dts-relative-module-declarations).

In essence, `declare module` statements with relative paths are not handled correctly: They are simply copied verbatim, which breaks the relative path. Take as an example a `config.ts` file that holds the following declaration:

```typescript
export interface Config {
  someProp: boolean;
}
```

and a `feature/extend-config.ts` file that extends that interface in a declared module:

```typescript
declare module "../config" {
  interface Config {
    featConfig: {
      featProp: number;
    }
  }
}
```

When running a program that references these two files through this plugin you will get:

```
export interface Config {
  someProp: boolean;
}

declare module "../config" {
  interface Config {
    featConfig: {
      featProp: number;
    }
  }
}
```

Here, the `declare module` no longer references the correct target. To fix this, the output should probably be either:

```typescript
export interface Config {
  someProp: boolean;
}

export interface Config {
  featConfig: {
    featProp: number;
  }
}
```

(note that all declarations or none must carry the `export` attribute), or:

```typescript
export interface Config {
  someProp: boolean;
}

declare module "./index" {
  interface Config {
    featConfig: {
      featProp: number;
    }
  }
}
```

Where “index” is the name of the output file. The latter would, of course, be difficult to do when writing to STDOUT as there is no output file name in such a situation.

This issue surfaced when I tried to migrate a project with libraries to Angular 20 which now uses this plugin. A downstream issue exists at ng-packagr/ng-packagr#3085.

I tried my best to provide a PR for either of the described approaches, but failed miserably for lack of understanding of how rollup works internally.

schoel-bis:
This fixes #345.

## Problem

Module declarations are just copied into the output file verbatim. This breaks for module declarations with relative paths. For example:

```typescript
declare module "../config" {
  interface Config {
    foo: number;
  }
}
```

This will just simply be copied, although the relative module path will no longer exist after running this plugin.

This breaks cross-module type extensions with relative paths, i.e. extending the pre-existing `Config` interface from `../config.ts` in the case above.

## Solution

To solve this, a new fixer is introduced that rewrites all relative module declarations in such a way, that they reference the output file itself, i.e. if the output file were called `index.d.ts`, the above would be rewritten as:

```typescript
declare module "./index" {
  interface Config {
    foo: number;
  }
}
```

Now, the pre-existing interface taken from `../config.ts` will be properly extended again.

### DIFF
--- a/src/transform/ModuleDeclarationFixer.ts
+++ b/src/transform/ModuleDeclarationFixer.ts
@@ -1,0 +1,56 @@
+import ts from "typescript";
+import MagicString from "magic-string";
+import { parse } from "../helpers.js";
+
+export class RelativeModuleDeclarationFixer {
+  private sourcemap: boolean;
+  private DEBUG: boolean;
+  private relativeModuleDeclarations: ts.ModuleDeclaration[];
+  private source: ts.SourceFile;
+  private code: MagicString;
+  private name: string;
+
+  constructor(fileName: string, code: MagicString, sourcemap: boolean, name?: string) {
+    this.sourcemap = sourcemap;
+    this.DEBUG = !!process.env.DTS_EXPORTS_FIXER_DEBUG;
+    this.relativeModuleDeclarations = [];
+    this.source = parse(fileName, code.toString());
+    this.code = code;
+    this.name = name || "./index";
+  }
+
+  fix() {
+    this.analyze(this.source.statements);
+
+    for (const node of this.relativeModuleDeclarations) {
+      const start = node.getStart();
+      const end = node.getEnd();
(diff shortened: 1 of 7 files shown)


Swatinem (approved):

alan-agius4:
@Swatinem, do you mind releasing this fix when you have a moment? Thanks.

PR #351, opened 2025-08-25
refactor: remove unused value reference tracking from DeclarationScope, Transformer, TypeOnlyFixer, and index

dayongkr:
To solve the issue in #348  , we actually only needed `this.convertTypeArguments(node)`, but it included excessive behavior like typeof value processing.

This PR removes those unnecessary parts. While the existing code wouldn't cause problems, I think it would be better to keep modifications as concise as possible when maintaining this project.


Swatinem (approved):
nice, thanks

PR #352, opened 2025-08-28
feat(transform): add support for child imports and nesting in namespaces

privatenumber:
Added support for `import` alias declarations (e.g. `import Alias = Other.Type`) inside `declare namespace` blocks:
- Update `DeclarationScope` to correctly hoist and resolve references from child imports within a namespace.
- New recursive pre-processing step to correctly handle arbitrarily nested namespaces.
- Refactor of the modifier fixing logic to be context-aware, preventing regressions in `declare module` and `declare global` blocks by applying different rules to top-level vs nested declarations.

Added new tests:

- `namespace-child-import`
	This is the foundational "happy path" test. It confirms the most basic requirement: a simple `import =` alias inside a single namespace is correctly rewritten to `type =`.

- `namespace-nested-shadowing`

	This is a complex scenario test. It validates three critical behaviors at once:
	1. The logic is recursive (handles nesting).
	2. The scope analysis correctly handles an alias shadowing an outer type.
	3. Tree-shaking correctly removes the unused (shadowed) type.

- `namespace-function-alias`

	This covers the specific edge case from the [`node-fetch` types](https://app.unpkg.com/@types/node-fetch@2.6.13/files/index.d.ts#L235). It asserts that an import alias pointing to a local identifier is correctly preserved, preventing the bug we fixed.

- `namespace-nested-export`

	This test confirms that export keywords on members inside a nested namespace are correctly preserved, which was a key part of the regression fix.

### DIFF
--- a/tests/testcases/namespace-child-import/meta.js
+++ b/tests/testcases/namespace-child-import/meta.js
@@ -1,0 +1,6 @@
+// @ts-check
+/** @type {import('../../testcases').Meta} */
+export default {
+  options: {},
+  rollupOptions: {},
+};
--- a/src/transform/DeclarationScope.ts
+++ b/src/transform/DeclarationScope.ts
@@ -373,6 +373,27 @@ export class DeclarationScope {
         }
         continue;
       }
+      if (ts.isImportDeclaration(stmt)) {
+        if (stmt.importClause) {
+          if (stmt.importClause.name) {
+            this.pushTypeVariable(stmt.importClause.name);
+          }
+          if (stmt.importClause.namedBindings) {
+            if (ts.isNamespaceImport(stmt.importClause.namedBindings)) {
+              this.pushTypeVariable(stmt.importClause.namedBindings.name);
+            } else {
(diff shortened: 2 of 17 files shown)


tests/testcases/namespace-child-import/meta.js
Swatinem, line 1: I think you can remove this meta file if you are not overriding any options.

Swatinem (approved):
nice, thank you!

typhonrt:
Greets @privatenumber / @Swatinem, 

I'd like to follow up with this PR for the reasons stated above. I do think there is a discussion to be had regarding the `export import` to `export type` conversion. IMHO no conversion should occur as `export import` is valid and required for interfaces being re-exported. If there are situations where this conversion is necessary the section in `preprocess.ts` should accommodate this aspect if it can be automatically detected otherwise an external configuration option is reasonable. By default, I think the types being bundled should not be altered in this case.

Otherwise everything else in this PR is great. Glad to hear the exact reasoning / discussion for the automatic conversion. I'll open a fresh issue next week / Monday if discussion is being missed here.

PR #360, opened 2025-10-16
fix(transform): preserve export keywords on nested namespaces

### LINKED ISSUE #359: [Code Review] Additional concerns about recent namespace handling changes (not released)
Greets @privatenumber,

I really appreciate your contributions to `rollup-plugin-dts`. 

I have continued to review your most recent changes and there seems to be one additional problem and one "annoyance" with PR #352. One of them is a definite problem that needs to potentially be addressed and another is an "annoyance" or likely unnecessary change in output types that likely can also be omitted. I will first discuss the problematic code change which removes the `export` keyword from nested namespaces followed by the "annoyance" which is likely an unnecessary invocation of `duplicateExports` on the nested namespace.

I am testing with the latest unpublished `rollup-plugin-dts` main repo as of (10/15/25) that has your changes incorporated. I am using NPM link to my complex project w/ nested namespaces / types. 

---

In particular lines [596-605](https://github.com/Swatinem/rollup-plugin-dts/blob/master/src/transform/preprocess.ts#L596-L605). This block of code removes the `export` keyword from nested namespaces. This effectively alters public scope availability of the nested namespaces making the nested namespace private. This affects downstream consumption for both types usage and API documentation purposes.

I will provide several picts below showing the original source types, the generated types before / after your changes. The following images will capture both the concerns being discussed.

Original source types:

Image

----

Original generated types (rollup-plugin-dts) prior to your changes (with my local workaround for namespaces [discussed here](https://github.com/Swatinem/rollup-plugin-dts/pull/352#issuecomment-3283535379)):

Image

----

Generated types after `PR 352`:

- The `red` line indicates the missing `export` keyword.
- The `blue` underline indicates unnecessary duplication of export names.

Image

----

The result in downstream usage of the nested namespace without the `export` keyword shows the breakage since it is no longer publicly accessible:

- This is from a consuming project using `TJSPosition`. 

Image

----

Original API docs before the change shows the `API` namespace:

Image

----

API docs after `PR 352` missing the `API` namespace due to it not being exported / public:

Image

----

I am not sure why I didn't catch the above sooner in my initial code review of your changes. I might have just stopped short and reported my initial findings regarding the `export import` keyword change for exported interfaces without finding these issues as well.

Removing lines [596-605](https://github.com/Swatinem/rollup-plugin-dts/blob/master/src/transform/preprocess.ts#L596-L605) will cause the tests you added to fail, but that is only because the test results expect the `export` keyword removed. 

IMHO this change to types should not occur as it changes intention and scoping of the original types for nested namespaces. 

----

Regarding the unnecessary invocation of `duplicateExports` on [lines 26-28](https://github.com/Swatinem/rollup-plugin-dts/blob/master/src/transform/preprocess.ts#L26-L28). I believe that this is entirely superfluous and can simply be removed. Removing it doesn't cause any tests to fail.  Again I believe original types should not be unnecessarily modified.

----

Almost there! I really appreciate your PRs and efforts to get this functionality rock solid and I hope you don't mind my feedback. I have a complex series of projects that use a lot of nested namespaces. 

It should be noted as of (10/15/25) when I run the tests as is `testcases/externals-link` fails without any modifications to the current source code in the main repo, so this is likely an unrelated failure in regard to the proposed changes above. 

I believe lines `26-28` and `596-605` can be safely removed from `preprocess.ts`.

Glad to have more discussion / clarification.

privatenumber:
Fixes #359

Removes two problematic code sections from PR #352:

1. **Lines 26-28**: Unnecessary `duplicateExports()` call that caused superfluous export name duplication in namespace bodies

2. **Lines 596-605**: Incorrect removal of export keywords from nested namespaces, which changed their accessibility from public to private

## Problem

The previous implementation was removing `export` keywords from nested namespaces, which:
- Changed nested namespaces from public to private scope
- Broke downstream consumption where nested namespaces were accessed
- Broke API documentation which relied on exported namespaces  
- Violated the plugin's philosophy: bundle declarations, not transform semantics

## Changes

- Removed unnecessary `duplicateExports()` invocation in `preProcessNamespaceBody()`
- Removed the `else` block in `fixModifiers()` that stripped `export` from nested namespaces
- Updated test expectations to preserve `export` keywords on nested namespaces

### DIFF
--- a/src/transform/preprocess.ts
+++ b/src/transform/preprocess.ts
@@ -23,10 +23,6 @@ function preProcessNamespaceBody(body: ts.ModuleBlock, code: MagicString, source
     // Safely call the new context-aware function on all children
     fixModifiers(code, stmt);
 
-    if (ts.isModuleDeclaration(stmt)) {
-      duplicateExports(code, stmt);
-    }
-
     // Recurse for nested namespaces
     if (ts.isModuleDeclaration(stmt) && stmt.body && ts.isModuleBlock(stmt.body)) {
       preProcessNamespaceBody(stmt.body, code, sourceFile);
@@ -593,16 +589,8 @@ function fixModifiers(code: MagicString, node: ts.Node) {
     if (needsDeclare && !hasDeclare) {
       code.appendRight(node.getStart(), "declare ");
     }
-  } else {
-    // For statements inside a namespace/module, only remove `export` from a nested namespace
-    // Leave all other members (vars, types, etc.) untouched
-    if (ts.isModuleDeclaration(node)) {
-      const exportKeyword = node.modifiers?.find((mod) => mod.kind === ts.SyntaxKind.ExportKeyword);
-      if (exportKeyword) {
-        code.remove(exportKeyword.getStart(), exportKeyword.getEnd() + 1);
-      }
-    }
   }
+  // For statements inside namespaces, preserve all modifiers (including export)
 }
 
 function duplicateExports(code: MagicString, module: ts.ModuleDeclaration) {
--- a/tests/testcases/namespace-nested-export/expected.d.ts
+++ b/tests/testcases/namespace-nested-export/expected.d.ts
@@ -2,7 +2,7 @@ interface External {
   id: string;
 }
 declare namespace Container {
-  namespace Nested {
+  export namespace Nested {
     // Import inside nested namespace with export keyword preserved
     import Local = External;
     export type FinalType = Local;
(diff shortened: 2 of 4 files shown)


typhonrt:
I have tested this PR in the real world against a complex project and the results are 100% great! Thanks again @privatenumber!

@Swatinem I recommend a merge for this PR and it should finish this round of namespace bundling improvements. This is significant new capabilities, so if possible please consider a new package release; perhaps `6.3.0` semver. Once this is publicly available it will allow me to do a major release of my OSS tooling that uses `rollup-plugin-dts`, so certainly anticipated.

PR #383, opened 2026-04-04
fix: rewrite shared chunk imports for TS2742 portability

privatenumber:
## Problem
`TS2742` shows up downstream when TypeScript has to name a public type through a private declaration path.

A small example looks like this:

```ts
// entry-a.d.ts
import { Shared } from "./shared.js";
export declare const a: Shared;

// entry-b.d.ts
export type { Shared } from "./shared.js";
```

A consumer can write:

```ts
import { a } from "lib/entry-a";
export const a2 = a;
```

and still get `TS2742`, because TypeScript does not go looking for some other public entry that happens to re-export `Shared`. It keeps following the declaration path it already has, which still points at the private shared chunk.

That is the case this PR is about. It is not a blanket "shared chunks are broken" issue. Some shapes are already portable, and some shapes are package-author problems that the plugin should not try to auto-fix.

For related context:
- [TypeScript issue: TS2742 with sibling re-export paths](https://github.com/microsoft/TypeScript/issues/60913)
- [TypeScript issue: TS2742 error-message DX](https://github.com/microsoft/TypeScript/issues/53764)

## What This PR Changes
This PR adds a narrow rewrite for the real plugin-side case.

When an entry leaks a type through a private shared chunk, and another public entry already re-exports that same type, the plugin rewrites the leaking entry to import through the public host entry instead.

So this:

```ts
// before
import { Shared } from "./shared.js";
export declare const a: Shared;
```

becomes this:

```ts
// after
import { Shared } from "./entry-b.js";
export declare const a: Shared;
```

That gives downstream TypeScript a public naming path such as:

```ts
export declare const a2: import("lib/entry-b").Shared;
```

The PR also adds a downstream consumer test harness so these cases are tested the way users actually hit them: by compiling a fake consumer package with `tsc --declaration`, then sanity-checking the emitted `.d.ts` again.

On top of the original repro, the suite now covers renamed host re-exports, `export *` hosts, multi-host selection, and mixed symbol visibility. For the `no-reexport` case, the plugin now emits a warning instead of silently leaving users with an unexplained downstream failure.

## Why This Solution
This is the smallest fix that improves real user experience without changing package API.

It works because it reuses a public path that already exists. It does not invent exports, merge hidden symbols into entry points, or flatten every shared chunk into an entry.

That matters for two reasons.

First, not every shared chunk is a bug. If an entry already exports the shared type publicly, downstream TypeScript can often name through that entry just fine. Rewriting everything would be unnecessary churn.

Second, when no public export exists, the plugin should not guess what the package API ought to be. Creating new exports would be a behavior change for the library, not just a portability fix.

## What Else We Considered
We considered treating every shared chunk as a plugin bug and always merging or rerouting it into an entry. The experiments did not support that. Some cases are already portable today, so that approach would solve the wrong problem and change more output than necessary.

We also considered auto-fixing the `no-reexport` case. That was rejected because it would create new public API surface. In that shape, the right move is guidance, not mutation, so this PR adds a warning instead.

We also verified that a sibling re-export alone is not enough. TypeScript does not discover it by itself; the declaration graph has to be rewritten explicitly. That is why the fix happens in the plugin output rather than in test expectations or docs alone.

## Testing
`npm test`

### DIFF
--- a/src/transform/sharedChunkPortability.ts
+++ b/src/transform/sharedChunkPortability.ts
@@ -1,0 +1,433 @@
+import * as path from "node:path";
+import type { OutputBundle, OutputChunk, SourceMap } from "rollup";
+import remapping from "@jridgewell/remapping";
+import type { RawSourceMap } from "@jridgewell/remapping";
+import ts from "typescript";
+import MagicString from "magic-string";
+import { parse } from "../helpers.js";
+
+type ChunkGraphInfo = {
+  exports: string[];
+  isEntry: boolean;
+};
+
+type ImportSpecifierInfo = {
+  importedName: string;
+  isTypeOnly: boolean;
+  localName: string;
+};
+
+type ImportStatementInfo = {
+  isTypeOnly: boolean;
+  moduleSpecifier: string;
+  quote: "'" | '"';
+  specifiers: ImportSpecifierInfo[];
+  statement: ts.ImportDeclaration;
+};
+
+type ImportBinding = {
+  importedName: string;
+  isTypeOnly: boolean;
+  moduleSpecifier: string;
+};
+
+type HostExportRoute = {
+  exportedName: string;
+  isTypeOnly: boolean;
+};
+
+type ChunkAnalysis = {
+  hostRoutesByModule: Map<string, Map<string, HostExportRoute[]>>;
+  importStatements: ImportStatementInfo[];
+};
+
+export function rewritePortableSharedChunkImportsInBundle(
+  bundle: OutputBundle,
+  warn: (message: string) => void,
+) {
+  const chunks = Object.values(bundle).filter(isOutputChunk);
+  const sharedChunks = chunks.filter((chunk) => !chunk.isEntry);
+  if (sharedChunks.length === 0) return;
+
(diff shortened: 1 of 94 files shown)


src/transform/sharedChunkPortability.ts
```diff
@@ -0,0 +1,433 @@
...
+          keptSpecifiers.push(specifier);
+          unresolvedSymbols.add(specifier.localName);
+          continue;
+        }
+
+        const hostSpecifier = getChunkImportSpecifier(chunk.fileName, hostCandidate.chunk.fileName);
+        const hostImportSpecifiers = rewrittenSpecifiers.get(hostSpecifier) || [];
+        hostImportSpecifiers.push({
+          importedName: hostCandidate.exportedName,
+          isTypeOnly: specifier.isTypeOnly || hostCandidate.isTypeOnly,
```
Swatinem, line 107: wouldn’t this mean that we could end up with a `type`-only import in case the `specifier` is actually a normal import and expects a runtime value to be imported?

Swatinem (approved):
thanks for your patience.
looks good, apart from a question that I have.

quite interesting that we have to end up reimplementing a lot of what I believe rollup does internally anyway.

PR #384, opened 2026-05-03
fix: avoid duplicate type modifier in exports

privatenumber:
## Summary
- Fixes a rewrite bug where namespace exports like `export { type T }` became invalid `export type { type T }`.
- When grouping type-only exports under `export type { ... }`, the fixer now prints only the export binding, e.g. `T` or `U as V`.
- Adds a regression fixture for plain and aliased namespace type exports.

## Test
- `npm test -- namespace-type-export-specifier`
- `npm test`


Swatinem (approved):
thanks.

I haven’t forgotten about your other PR. Its quite large and I haven’t found enough motivation yet to look at it thoroughly.

PR #386, opened 2026-05-19
fix: correctly resolve string literals in export specifiers

cueaz:
### Description

Fixes the resolution of ES2022 string literal export names (e.g., `export { b as "bb" }`).

Currently, Rollup fails to link identifiers exported as string literals, resulting in missing exports and `[undefined]` properties in the bundled output. This occurs because `createIdentifier` uses `node.getText()`, which preserves the surrounding quotes for `ts.StringLiteral` nodes (e.g., `'"bb"'`), causing a mismatch in Rollup's linker.

Updating this to use `node.text` passes the unquoted identifier value, resolving the issue.

##### `mod.d.ts`

```ts
declare const a: "a";
declare const b: "b";
declare const c: "c";

export { a as aa, b as "bb", c as "c c" };
```

##### `index.d.ts`

```ts
import * as mod from "./mod";
import { "c c" as cc } from "./mod";

declare class Test {
  [mod.aa]: string;
  [mod.bb]: string;
  [cc]: string;
}

export { Test };
```

##### Actual Output (Failing):

```ts
declare const a: "a";
declare const c: "c";
declare class Test {
  [a]: string;
  [undefined]: string;
  [c]: string;
}
export { Test };
```

##### `expected.d.ts`

```ts
declare const a: "a";
declare const b: "b";
declare const c: "c";
declare class Test {
  [a]: string;
  [b]: string;
  [c]: string;
}
export { Test };
```

### DIFF
--- a/src/transform/astHelpers.ts
+++ b/src/transform/astHelpers.ts
@@ -41,7 +41,7 @@ export function createIdentifier(node: ts.Identifier | ts.StringLiteral): ESTree
   return withStartEnd(
     {
       type: "Identifier",
-      name: node.getText(),
+      name: node.text,
     },
     node,
   );
--- a/tests/testcases/string-literal-export-names/expected.d.ts
+++ b/tests/testcases/string-literal-export-names/expected.d.ts
@@ -1,0 +1,9 @@
+declare const a: "a";
+declare const b: "b";
+declare const c: "c";
+declare class Test {
+  [a]: string;
+  [b]: string;
+  [c]: string;
+}
+export { Test };
--- a/tests/testcases/string-literal-export-names/index.d.ts
+++ b/tests/testcases/string-literal-export-names/index.d.ts
@@ -1,0 +1,10 @@
+import * as mod from "./mod";
+import { "c c" as cc } from "./mod";
+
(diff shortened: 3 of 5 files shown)


Swatinem (approved):
I’m actually surprised that this works, and doesn’t require the AST node to be a `Literal` instead.

But okay, I guess this is simpler so why not :-)

Swatinem:
well, CI says the syntax is not supported.

Does it depend on TS6, which I think we haven’t updated to yet, or does it need a specific `tsconfig`, in which case you would need to add that to a `meta` file. There are a bunch of those among the testcases for inspiration.

cueaz:
Ah, you're right. It requires TS 5.6 for arbitrary module identifiers. I'll add the meta.js file.

For the identifier part, I also worried about invalid names, but followed `createIdentifier` since it accepts `StringLiteral`. It’s interesting that Rollup links them without errors.

Swatinem (approved):

Swatinem:
thanks for the fix.
even though without a TS update, the tests are effectively disabled, the actual fix is trivial enough to not cause problems.

PR #392, opened 2026-07-02
ci: renovate CI workflow

igordanchenko:
## Summary

Small CI maintenance pass on `.github/workflows/ci.yml`:

- **Replace the `typescript@beta` compat run with `typescript@5`.** The beta step was there to smoke-test the then-unreleased TypeScript 6. Now that the pinned `typescript` devDependency *is* 6.x (`6.0.3`), that run just re-tests the same major as the main build. Meanwhile the **5.x** peer range — by far the most widely deployed major — was going completely untested. Pointing the step at `typescript@5` closes that gap.
- **Enable the built-in npm cache** on `actions/setup-node` (`cache: npm`) to speed up `npm ci`.

## Compat matrix, before vs. after

The peer range is `typescript@^4.5 || ^5.0 || ^6.0`. The extra install-and-retest steps at the bottom of the job exercise the edges of that range:

| Run              | Before        | After         |
| ---------------- | ------------- | ------------- |
| Main build/test  | TS 6.0.3      | TS 6.0.3      |
| Compat run       | TS beta (6.x) | **TS 5**      |
| Minimum-peer run | TS 4.5        | TS 4.5        |

## Deliberately out of scope

Noticed while reviewing, but **not** touched in this PR — happy to open follow-ups if wanted:

- **The Node matrix uses `latest`**, which resolves to the newest *released* Node line — currently **Node 26** (`v26.4.0`), a **Current** (non-LTS) release that won't reach Active LTS until ~October 2026. Both the test matrix and, more notably, the **codecov upload** (`if: ... matrix.node == 'latest'`) are pinned to it. Was that intentional, or should the coverage run be anchored to an LTS? No change made pending your call. Node 22/24 LTS lines are also currently untested (only `20` and `latest` run).
- **The minimum-peer run installs `rollup@3.0`**, but `peerDependencies` declares `rollup@^3.29.4` — so CI tests a version below the stated floor.

## Test plan

- CI green across the matrix.
- Confirm the `typescript@5` job resolves a 5.x version and the test suite passes against it.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -18,6 +18,7 @@ jobs:
       - uses: actions/setup-node@v6
         with:
           node-version: ${{ matrix.node }}
+          cache: npm
       - run: npm ci
       - run: npm test
       # the codecov uploader only runs on unix,
@@ -29,8 +30,8 @@ jobs:
         continue-on-error: true
         run: bash <(curl -s https://codecov.io/bash) -t ${{secrets.CODECOV_TOKEN}} -B ${{ github.ref }} -f coverage/coverage-final.json
 
-      # test TypeScript 6 beta
-      - run: npm install typescript@beta
+      # test against TypeScript 5
+      - run: npm install typescript@5
       # aka `npm test` without the `pretest/build`
       - run: node .build/tests/index.js
 


igordanchenko:
Heads up: the red CI here isn't from these changes — `master` itself is currently failing. The five fixture failures were introduced by #383 (merged against stale checks) and are already fixed in #391 (test-only). This PR only touches `ci.yml`, so I'll rebase once #391 lands and CI should go green.

.github/workflows/ci.yml
```diff
@@ -29,8 +30,8 @@ jobs:
         continue-on-error: true
         run: bash <(curl -s https://codecov.io/bash) -t ${{secrets.CODECOV_TOKEN}} -B ${{ github.ref }} -f coverage/coverage-final.json
 
-      # test TypeScript 6 beta
-      - run: npm install typescript@beta
+      # test against TypeScript 5
+      - run: npm install typescript@5
```
Swatinem, line 34: IMO, testing the two edges (latest, and oldest) is quite adequate. I don’t think there is enough value to test the LTS / most widely used version. On that note, your AI mentioned that we test a lower rollup version than the peer dependency hints at, so maybe lowering the peer dependency is worth doing. Or not, as I doubt anyone is still using that.

Might as bump that a bit higher and move on.

igordanchenko: Usage of 3.0.0 – 3.29.3 appears to be negligible.

| Range           | Weekly downloads | Share  |
| --------------- | ---------------- | ------ |
| 4.x             | 113,849,098      | 87.98% |
| 3.29.4 – 3.30.0 | 3,910,437        | 3.02%  |
| 3.0.0 – 3.29.3  | 634,641          | 0.49%  |

While there's probably not a lot of value in extending support to the 3.0.0 – 3.29.3 range (you have to wonder what percentage of those downloads are just CI runs), I got curious why ^3.29.4 is the current peer dependency range. As it turns out, it's courtesy of Renovate bot. It looks like you intentionally set the peer dependency range to ^3.0 at least twice, but Renovate kept bumping it back up. 

Interestingly, `renovate.json` already has a rule meant to prevent this ("Ignore node and peerDependencies"), but it doesn't work as intended: all `match*` conditions within a `packageRule` are ANDed together, so `matchPackageNames: ["node"]` limits the whole rule to the `node` package, and `rollup` in `peerDependencies` is never matched. Splitting it into two rules would fix it:

```json
"packageRules": [
  {
    "description": "Ignore Node engine updates",
    "matchManagers": ["npm"],
    "matchDepTypes": ["engines"],
    "matchDepNames": ["node"],
    "enabled": false
  },
  {
    "description": "Ignore peerDependencies",
    "matchManagers": ["npm"],
    "matchDepTypes": ["peerDependencies"],
    "enabled": false
  }
]
```

With that said, I'd lower the peer dependency range to ^3, since it's already CI-tested and matches your original intent.

As for the Node version, it caught my eye because `latest` (26) is not a production-grade release yet. Per the Node.js [release recommendations](https://nodejs.org/en/about/previous-releases): _"Production applications should only use Active LTS or Maintenance LTS releases."_

igordanchenko: I've submitted a separate PR (#396) to address the rollup peer dependency range and the Renovate config.

igordanchenko: @Swatinem, did I address your concerns or should I close this PR?

Swatinem (approved):

PR #394, opened 2026-07-03
fix: harden relative `declare module` specifier rewriting

### LINKED ISSUE #393: Relative `declare module` rewriting produces wrong or corrupted output in several scenarios
### Checklist

## Environment

- rollup-plugin-dts 6.4.1
- Reproduced with rollup 4.61 / TypeScript 6.0 (and, where applicable, at the minimum supported peers rollup 3.0 / TypeScript 4.5)

## Background

#350 and #372 rewrite relative `declare module './x'` specifiers in bundled output so augmentations point at the output chunk containing their target. Those PRs deliberately kept scope to the most pressing use cases; a follow-up audit of the surrounding edge cases surfaced six scenarios worth fixing now that the core mechanism has proven itself. By their nature these are invisible to snapshot tests — string comparison can't tell whether an augmentation still *applies* for a consumer of the bundled output — so the fixes come with tsc-level assertions alongside the snapshots.

## 1. Unresolvable targets are rewritten to the current chunk, merging the augmentation into the wrong module

When the probe finds no chunk for the target (the file exists but was never imported, or the specifier points at a non-module asset), the specifier is rewritten to the current chunk's own name as a fallback.

```ts
// index.d.ts (entry)
export interface Config {
  base: string;
}
export declare function getConfig(): Config;

declare module "./orphan" {
  // orphan.d.ts exists on disk but is never imported
  interface Config {
    fromAugmentation: string;
  }
}
```

The output contains `declare module "./index" { interface Config { … } }` — a self-augmentation of the chunk. tsc **applies** that augmentation: the chunk's own exported `Config` silently gains a required `fromAugmentation` member, changing the public types of an unrelated interface.

Keeping the original specifier would be strictly safer: tsc silently ignores an unresolvable relative `declare module` inside a module `.d.ts`, so the augmentation stays dormant instead of corrupting the chunk's types.

## 2. Asset-style declarations are destroyed and create phantom exports

Same fallback, different victim:

```ts
declare module "./logo.svg" {
  const url: string;
  export { url };
}
```

becomes `declare module "./index" { const url: string; export { url }; }` — the ambient asset declaration is gone, and the chunk's types now claim a `url` export that does not exist at runtime. (Wildcard patterns like `declare module "*.svg"` are unaffected; only specific relative specifiers hit this.)

## 3. Directory-index targets never resolve

The chunk lookup probes `basePath + extension` but never `basePath + "/index" + extension`, so `declare module './utils'` backed by `utils/index.d.ts` always misses — *even when that module is bundled into the very chunk being processed* — and falls into the issue-1 rewrite.

## 4. `.tsx` / `.jsx` module ids never resolve

Rollup module ids keep the original source path (`resolveId` returns the resolved file name verbatim), and the plugin accepts `.tsx`/`.jsx` inputs, but the probe's extension list has no JSX variants. `declare module './Component'` targeting a bundled `Component.tsx` always misses.

## 5. Renamed chunk exports silently retarget the augmentation to the wrong type

The rewrite retargets the *specifier* but never the names inside the augmentation body, while module augmentation matches by the target module's **exported** names. When bundling renames an export — aliased re-exports (`export type { Foo$1 as FooA, Foo as FooB }`) or deconfliction when two modules in one chunk both export `Foo` — the two desynchronize:

- the intended type never receives the augmentation (consumer sees TS2339 on the augmented member), and
- worse, if another module's same-named export kept the original name, the augmentation **silently merges into that unrelated type**.

Verified with tsc: given a chunk exporting both `Foo` (module A) and `Foo$1` (module B, renamed), an augmentation written for B's `Foo` adds its members to A's `Foo` and leaves `Foo$1` untouched.

## 6. Corrupted output when a type-only fix precedes a `declare module`

`ModuleDeclarationFixer` receives `TypeOnlyFixer`'s already-edited MagicString, parses `toString()` (post-edit text) to compute node positions, then calls `overwrite()` — whose offsets MagicString interprets in **original**-string coordinates. Any length-changing type-only edit before the declare module (e.g. `export { Foo }` → `export type { Foo }` inside a namespace) shifts every later offset:

```
declare module "./ot"./index.d" nterface Extra {
```

— leaked resolution marker, mangled specifier, eaten body start, in the published `.d.ts`.

## Known limitation (out of scope, for the record)

tsconfig-paths aliases (`declare module "@/foo"`) are not rewritten — they cannot be resolved lexically. This is a deliberate scope restriction (bare package names must stay untouched); a design for supporting them is sketched in the fix PR.

## Fix

PR to follow, addressing all six: keep the original specifier on lookup miss (issues 1–2), probe directory indexes and JSX extensions (3–4), guard the rewrite on the target chunk exporting the augmented names unchanged (5), and align MagicString coordinate spaces (6) — with tsc-level tests asserting augmentations both apply after a rewrite and stay dormant when the specifier is kept.

igordanchenko:
## Summary

Follow-up to #350 / #372, which deliberately kept scope to the most pressing use cases. This PR hardens the mechanism against the edge cases around that scope — six scenarios detailed in the linked issue — across four behavior commits plus one documentation commit, and adds tsc-level assertions that augmentations actually apply (or stay dormant) for a consumer of the bundled output.

Fixes #393

## Commits

Each commit is independently green and reviewable on its own.

### 1. `fix: keep original specifier for unresolvable relative declare module`

When the target cannot be matched to any output chunk, the specifier is now preserved (only the internal `dts-resolved:` marker is stripped) instead of being rewritten to the current chunk. A dangling relative augmentation in a module `.d.ts` is silently ignored by tsc, so the augmentation stays dormant — strictly safer than the previous behavior, where a same-named type exported by the current chunk silently gained the augmentation's members, and asset-style declarations turned into phantom exports on the chunk.

The miss warning now includes both the original specifier and the resolved absolute path, keeping same-named specifiers from different directories distinguishable:

```
declare module "./orphan" (/abs/src/orphan) could not be resolved to any output chunk, keeping the original specifier
```

### 2. `fix: resolve directory index and tsx targets of relative declare module`

The chunk lookup now probes `basePath + "/index"` variants and includes `.tsx`/`.jsx` in the extension list (Rollup module ids keep the original source path). Exact-path matches still win over index probing, matching Node resolution precedence. Targets like `./utils` → `utils/index.d.ts` and `./Component` → `Component.tsx` are rewritten to the correct chunk instead of warning and (pre-PR) being rewritten to the wrong one.

### 3. `fix: skip declare module rewrite when augmented names are renamed`

Module augmentation matches by the target module's exported names. The specifier is now rewritten only when the target chunk exports **every name declared in the augmentation body unchanged**: still rendered by the target module (`modules[id].renderedExports`), present in `chunk.exports`, and with no `name$<digits>` deconfliction sibling. Otherwise the original specifier is kept and a warning is emitted:

```
declare module "./foo-b" (/abs/src/foo-b) augments names that the target chunk does not export unchanged, keeping the original specifier
```

This prevents the silent wrong-type merge: previously, when two modules in one chunk both exported `Foo`, an augmentation intended for the renamed one merged its members into the *other* module's `Foo`.

### 4. `fix: prevent corrupted output when type-only fixes precede declare module`

`ModuleDeclarationFixer` now parses the MagicString's `original` text instead of `toString()`, so node positions and `overwrite()` coordinates share one coordinate space regardless of edits already recorded by `TypeOnlyFixer`. Previously, a length-changing type-only edit before a `declare module` statement shifted every later offset and garbled the emitted file (leaked marker, mangled specifier). Type-only edits never overlap a module declaration header, so both fixers safely share one MagicString and one composed sourcemap.

### 5. `docs: document declare module rewrite scope and alias limitation`

Class-level documentation of the marker → probe → rewrite/keep mechanism, and why the specifier filter is deliberately limited to `./`/`../`: bare package names must be preserved verbatim, and tsconfig-paths aliases cannot be resolved lexically. Includes the recipe for adding alias support later (paths-pattern-gated `ts.resolveModuleName` in preprocess, skipping `isExternalLibraryImport` results).

## Behavior changes

| Scenario | Before | After |
| --- | --- | --- |
| Target not in any chunk (orphan, asset) | Rewritten to current chunk → wrong-module merge / phantom exports | Original specifier kept (dormant) + warning |
| Target is `dir/index.*` | Miss → rewritten to current chunk | Rewritten to the correct chunk |
| Target is a `.tsx`/`.jsx` module | Miss → rewritten to current chunk | Rewritten to the correct chunk |
| Augmented name renamed/aliased/dropped by the chunk | Rewritten → silent wrong-type merge or dead augmentation | Original specifier kept (dormant) + warning |
| Type-only fix before `declare module` | Corrupted output text | Clean output |

## Tests

Five new testcases (`declare-relative-module-{unresolved,index,tsx,renamed,type-only}`), including downstream consumers that compile against the bundle packaged as a node module:

- `-index` proves a rewritten augmentation **applies**: the consumer compiles `u.augmented` (moduleResolution `Bundler`).
- `-unresolved` and `-renamed` prove kept augmentations stay **dormant** with no collateral merge, via `@ts-expect-error` assertions that fail (TS2578) if the augmentation ever starts applying — a direct regression trip-wire for the pre-PR behavior.
- `-type-only` pins the coordinate-space fix with a length-changing type-only edit before the declare module; reverting the fix garbles the snapshot.

Full suite verified on current deps and at the minimum supported peers (`typescript@4.5` + `rollup@3.0`, matching the CI compat step). Four of the five testcases run at the minimum peers too; `-renamed` is gated `rollupVersion: "4.0.0"` because its snapshot encodes Rollup 4 chunk hosting and alias names — early Rollup 3 emits an equivalent but differently-shaped chunk (the fixer behavior is identical there; verified via its downstream consumer, which passes on 3.x).

## Known accepted residuals (documented in code/tests)

- A kept specifier re-resolves relative to the emitted chunk, so dormancy is not *guaranteed* if a file coincidentally exists at that path next to the chunk — still strictly narrower risk than the previous always-rewrite.
- The rename guard works from chunk metadata; a deconflicted name whose rename is unexported while another module's same-named export kept the original name is invisible to it. Requires internal-only usage of the renamed type; strictly rarer than the pre-guard behavior.
- tsconfig-paths aliases remain unsupported (documented limitation with an implementation sketch in the code).

### DIFF
--- a/src/transform/ModuleDeclarationFixer.ts
+++ b/src/transform/ModuleDeclarationFixer.ts
@@ -51,6 +51,35 @@ function getAugmentedNames(body: ts.ModuleBlock) {
   return names;
 }
 
+/**
+ * Rewrites relative `declare module './x'` specifiers in bundled output so they point
+ * at the chunk containing the target module, keeping module augmentations reachable
+ * for consumers.
+ *
+ * preprocess resolves each relative specifier lexically against its source file's
+ * directory and stamps it with a `dts-resolved:` marker comment that rides through
+ * bundling attached to its declaration; this fixer decodes the marker, locates the
+ * target chunk via Rollup's chunk metadata, and overwrites the specifier with a
+ * chunk-relative path.
+ *
+ * The rewrite is skipped (original specifier kept + warning) when:
+ * - the target module is not part of any output chunk (orphaned files, asset-style
+ *   declarations) — a dangling specifier stays inert, while rewriting it would merge
+ *   the augmentation into the wrong module;
+ * - the target chunk does not export every augmented name unchanged (aliased
+ *   re-exports, `Foo` → `Foo$1` deconfliction) — augmentation matches by exported
+ *   name, so retargeting the specifier alone would desynchronize the two.
+ *
+ * Scope: only `./`/`../` specifiers are handled. Bare package names must be preserved
+ * verbatim, since consumers resolve them by name. tsconfig-paths aliases
+ * (`declare module "@/foo"`) are a known limitation — they cannot be resolved
+ * lexically. Supporting them would mean: in preprocess, for specifiers matching a
+ * `compilerOptions.paths` pattern (never for other bare names), resolve with
+ * `ts.resolveModuleName` against the declaring file's compiler options (synchronous,
+ * unlike the plugin context's `resolve`), skip `isExternalLibraryImport` results, and
+ * stamp the resolved file name; the extension probe below already bridges the
+ * remaining `.d.ts`-versus-`.ts` module id gap.
+ */
 export class ModuleDeclarationFixer {
   private code: MagicString;
   private sourcemap: boolean;
--- a/src/transform/ModuleDeclarationFixer.ts
+++ b/src/transform/ModuleDeclarationFixer.ts
@@ -105,9 +105,15 @@ export class ModuleDeclarationFixer {
     const jsExtMatch = absolutePath.match(/\.[cm]?js$/);
     const basePath = jsExtMatch ? absolutePath.slice(0, -jsExtMatch[0].length) : absolutePath;
 
-    // Try all file extensions that could appear as module IDs in Rollup's chunk metadata
-    const extensions = ["", ".d.ts", ".d.mts", ".d.cts", ".ts", ".mts", ".cts", ".js", ".mjs", ".cjs"];
-    const possiblePaths = extensions.map((ext) => basePath + ext);
+    // Try all file extensions that could appear as module IDs in Rollup's chunk metadata,
+    // probing both the path itself and a directory index
+    const extensions = ["", ".d.ts", ".d.mts", ".d.cts", ".ts", ".tsx", ".mts", ".cts", ".js", ".jsx", ".mjs", ".cjs"];
+    const possiblePaths: string[] = [];
+    for (const base of [basePath, `${basePath}/index`]) {
+      for (const ext of extensions) {
+        possiblePaths.push(base + ext);
+      }
+    }
 
     // Find which chunk contains this module
     for (const possiblePath of possiblePaths) {
--- a/src/transform/ModuleDeclarationFixer.ts
+++ b/src/transform/ModuleDeclarationFixer.ts
@@ -68,18 +68,25 @@ export class ModuleDeclarationFixer {
 
       const targetChunkName = this.getTargetChunkName(absolutePath);
 
-      const quote =
-        node.name.kind === ts.SyntaxKind.StringLiteral && "singleQuote" in node.name && node.name.singleQuote
-          ? "'"
-          : '"';
+      let specifier = node.name.getText();
+      if (targetChunkName === null) {
+        // Keep the original specifier; consumers resolve it relative to the emitted
+        // chunk, where it will typically dangle (a no-op augmentation) rather than
+        // merge into the wrong module the way a current-chunk rewrite would
+        this.warn(
+          `declare module ${specifier} (${absolutePath}) could not be resolved to any output chunk, keeping the original specifier`,
+        );
+      } else {
+        const quote =
+          node.name.kind === ts.SyntaxKind.StringLiteral && "singleQuote" in node.name && node.name.singleQuote
+            ? "'"
+            : '"';
+        specifier = `${quote}${targetChunkName}${quote}`;
+      }
 
       const cleanedBetween = stripResolvedModuleComment(textBetween);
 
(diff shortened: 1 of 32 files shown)


igordanchenko:
Heads up: the red CI here isn't from these changes — `master` itself is currently failing. The five fixture failures were introduced by #383 (merged against stale checks) and are already fixed in #391 (test-only). I verified locally that this branch fails the exact same five fixtures as `master`, and none of the new `declare-relative-module-*` tests are among them. I'll rebase once #391 lands and CI should go green.

Swatinem (approved):

PR #401, opened 2026-07-30
feat: support TypeScript 7.0

### LINKED ISSUE #395: TypeScript 7.0 support
## Context

TypeScript 7.0 (the native compiler) was released on July 8, 2026: 
https://devblogs.microsoft.com/typescript/announcing-typescript-7-0/

Since `rollup-plugin-dts` relies on the TypeScript compiler API at runtime (`ts.createProgram`, `ts.resolveModuleName`, AST traversal), this release affects the plugin directly: **TypeScript 7.0 does not ship a stable compiler API**. The main entry of the `typescript@7` package exports only `version` / `versionMajorMinor`. An early API does exist, but only under explicitly-unstable subpath exports (`typescript/unstable/sync`, `typescript/unstable/ast`, `typescript/unstable/ast/factory`, …), and the announcement states a new (and different) stable API is expected in TypeScript 7.1.

## Impact

- Projects that upgrade to `typescript@7` get no compiler API from their `typescript` package, so the plugin cannot create programs or resolve modules and fails at load time.
- The current `peerDependencies` range (`^4.5 || ^5.0 || ^6.0`) does not admit `7.x`.

## Interim path recommended by the TypeScript team

Microsoft publishes `@typescript/typescript6`, a compatibility package carrying the TypeScript 6.0 API (with a `tsc6` binary), and recommends npm aliases for tools that need both the native compiler and the JS API:

```json
{
  "devDependencies": {
    "@typescript/native": "npm:typescript@^7.0.2",
    "typescript": "npm:@typescript/typescript6@^6.0.2"
  }
}
```

## Proposal

- At runtime, load the API from the user's `typescript` package when it provides one (TS 4.5–6.x); when it doesn't (TS 7), fall back to `@typescript/typescript6`, and fail with an actionable error ("please install `@typescript/typescript6`") when neither is available. Declaration parsing/emit behavior stays on TS 6 semantics, which is unchanged for `.d.ts` output.
- Extend `peerDependencies` to `^4.5 || ^5 || ^6 || ^7`, adding `@typescript/typescript6` as an optional peer.
- Use the aliases above in `devDependencies` so the repo builds with the native compiler while source files keep their `typescript` types.
- Revisit once TypeScript 7.1+ ships the stable API (the current `unstable/*` exports are explicitly subject to change, so I'd avoid building against them).

Known limitation in the interim: the published `.d.ts` still references `typescript` for types like `ts.CompilerOptions`, so a TS7-only project type-checking its Rollup config may see missing-member errors until 7.1; the alias pattern above resolves that too.

I have this working on a branch (all tests green, including the downstream consumer tests now type-checked with the native 7.0 `tsc`) and can open a PR if you are okay with the above plan.

igordanchenko:
TypeScript 7.0 does not ship the JS compiler API, so the published bundle now loads the API at runtime: from the user's `typescript` package when it provides one (TypeScript 4.5 - 6.x), with a fallback to the official `@typescript/typescript6` compatibility package under `typescript@7`, and a guided error when neither is available. Most of the diff is the lockfile.

- `src/typescript-shim.mjs` is the runtime loader (~40 lines). Fully synchronous via `createRequire`, resolved once at module load, just like the previous static import. A small `resolveId` hook in `rollup.config.ts` substitutes it for `import ts from "typescript"` at bundle time, so source files are untouched and keep the real `typescript` import for its types; only the published bundle is redirected.
- `package.json` allows `typescript@7` in peer dependencies and adds `@typescript/typescript6` as an optional peer. Nothing changes for TS <= 6 users; declaring the optional peer also makes the fallback resolvable under pnpm's strict layout and Yarn PnP.
- `tests/shim.ts` is a new smoke test that runs the published bundles (both entry points) in a synthetic `node_modules`, covering all shim paths: TS with the API, TS 7 with the fallback, TS 7 without it, and no TS at all. The other tests import the plugin from source, so this is the only place the shim is actually executed.
- `README.md` documents the setup for `typescript@7` users, including a note that they need `skipLibCheck` when type-checking code that imports the plugin (the public types reference `ts.CompilerOptions`, which `typescript@7` does not provide).
- The repo itself now builds with the native TypeScript 7 compiler (via the `@typescript/native` npm alias), and the downstream tests type-check bundled output with it, while source type-checking stays on the TS 6 API types via the `typescript@npm:@typescript/typescript6` alias.

Verified beyond CI:

- `npm install` peer resolution against real `typescript@7.0.2`, and against the alias setup recommended in the TS release notes
- end-to-end `.d.ts` bundle through the fallback path, using the packed tarball in a fresh project

TypeScript 7.0 introduces no new syntax, so parsing and emitting declarations with the 6.0 API produces identical output. Once the new API ships in TS 7.1+, the shim can be deleted wholesale as part of that migration.

Closes #395

### DIFF
--- a/tests/shim.ts
+++ b/tests/shim.ts
@@ -31,6 +31,15 @@ if (typeof dts.default !== "function") {
 }
 `;
 
+const CONSUMER_CONFIG = `import dts from "rollup-plugin-dts";
+
+export default {
+  input: "src/index.d.ts",
+  output: { file: "out/index.d.ts", format: "es" },
+  plugins: [dts()],
+};
+`;
+
 // use junctions so directory links work on Windows without elevated privileges
 async function link(target: string, linkPath: string) {
   await fs.mkdir(path.dirname(linkPath), { recursive: true });
@@ -46,8 +55,12 @@ async function createFixture(tempRoot: string) {
   await fs.writeFile(path.join(tempRoot, "src", "index.d.ts"), SAMPLE_DTS);
   await fs.writeFile(path.join(tempRoot, "consumer.mjs"), CONSUMER_MJS);
   await fs.writeFile(path.join(tempRoot, "consumer.cjs"), CONSUMER_CJS);
+  await fs.writeFile(path.join(tempRoot, "rollup.config.mjs"), CONSUMER_CONFIG);
 
   await fs.mkdir(pluginDir, { recursive: true });
+  // dist must be copied, not linked: Node resolves modules to their real path,
+  // so a linked dist would make the shim resolve `typescript` from this repo's
+  // node_modules and defeat the fixture isolation
   await fs.cp(path.resolve("dist"), path.join(pluginDir, "dist"), { recursive: true });
   await fs.writeFile(
     path.join(pluginDir, "package.json"),
@@ -93,6 +106,14 @@ export default (t: Harness) => {
       assert.ok(fallback.stdout.includes("interface Foo"), fallback.stdout);
       const fallbackCjs = runNode(tempRoot, ["consumer.cjs"]);
       assert.strictEqual(fallbackCjs.status, 0, fallbackCjs.stderr);
+
+      // the rollup cli loads the config file through its own bundling step,
+      // so also cover that path importing the plugin
+      const rollupBin = path.join(tempRoot, "node_modules", "rollup", "dist", "bin", "rollup");
+      const cli = runNode(tempRoot, [rollupBin, "--config", "--silent"]);
+      assert.strictEqual(cli.status, 0, `${cli.stdout}${cli.stderr}`);
+      const cliOutput = await fs.readFile(path.join(tempRoot, "out", "index.d.ts"), "utf-8");
+      assert.ok(cliOutput.includes("interface Foo"), cliOutput);
 
       // typescript@7 without the fallback: guided error
       await fs.rm(ts6Link, { recursive: true, force: true });
--- a/tests/shim.ts
+++ b/tests/shim.ts
@@ -1,0 +1,112 @@
+import * as assert from "assert";
+import { spawnSync } from "child_process";
+import fs from "fs/promises";
+import * as os from "os";
+import * as path from "path";
+import { Harness } from "./utils.js";
+
+// The published bundles load the TypeScript compiler API through
+// src/typescript-shim.mjs, which the other tests never execute (they import
+// the plugin from source, where "typescript" resolves directly). Smoke-test
+// the shim against the actual dist bundles in a synthetic node_modules with
+// different `typescript` installations.
+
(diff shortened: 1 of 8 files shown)


tests/shim.ts
```diff
@@ -0,0 +1,112 @@
...
+  const pluginDir = path.join(tempRoot, "node_modules", "rollup-plugin-dts");
+
+  await fs.mkdir(path.join(tempRoot, "src"), { recursive: true });
+  await fs.writeFile(path.join(tempRoot, "package.json"), '{ "type": "module" }\n');
+  await fs.writeFile(path.join(tempRoot, "src", "index.d.ts"), SAMPLE_DTS);
+  await fs.writeFile(path.join(tempRoot, "consumer.mjs"), CONSUMER_MJS);
+  await fs.writeFile(path.join(tempRoot, "consumer.cjs"), CONSUMER_CJS);
+
+  await fs.mkdir(pluginDir, { recursive: true });
+  await fs.cp(path.resolve("dist"), path.join(pluginDir, "dist"), { recursive: true });
```
Swatinem, line 64: might as well use `link` for this as well, or is there a specific reason not to?

tests/shim.ts
```diff
@@ -0,0 +1,112 @@
...
+// the shim against the actual dist bundles in a synthetic node_modules with
+// different `typescript` installations.
+
+const SAMPLE_DTS = `export interface Foo {
+  bar: number;
+}
+export type Bar = Foo["bar"];
+`;
+
+const CONSUMER_MJS = `import { rollup } from "rollup";
```
Swatinem, line 20: Another interesting case would be to run the rollup cli with a `rollup.config.js` file. I remember rollup does some fancy things when loading its config file.

Swatinem (approved):

tests/shim.ts
igordanchenko: There is a reason: Node resolves modules to their real path when loading them, so with a linked `dist` the shim's `createRequire` would resolve `typescript` relative to this repo's actual `dist` directory and find the repo's own `node_modules/typescript`. That silently defeats the fixture isolation. The smoke test itself catches this: with `dist` linked, the missing-fallback scenarios stop failing and the assertions trip.

tests/shim.ts
igordanchenko: Good call, added. The fixture now also runs the `rollup` CLI with a `rollup.config.mjs` that imports the plugin (in the TS 7 + fallback scenario) and asserts the emitted output. The CLI's config bundling keeps bare imports external, so the plugin resolves from the fixture's `node_modules` the same way it does in a real project, shim included.

RainbowMango:
Great job! @Swatinem, and thank you for your effort. Just a question here: are you going to make a new tag for this?

